PR #1, opened 2025-08-11
Add qcom-preflight-checks

njjetha:
## Summary
qcom-preflight-checks calls a [reusable-workflow](https://github.com/qualcomm/qcom-reusable-workflows/blob/main/.github/workflows/qcom-preflight-checks-reusable-workflow.yml) that runs a series of preflight checks on your proposed contribution.

For more info check https://github.com/qualcomm/qcom-actions/blob/main/README.md


b49020 (approved):
LGTM

ricardosalveti:
Do we want the preflight checks on forked projects like this one? If so we should probably tune the configs for what is really relevant here.

b49020:
@ricardosalveti 

> Do we want the preflight checks on forked projects like this one?

I am not really sure either. Since these forked projects will closely follow upstream tree, maybe we can follow what we already do for the kernel repo. @idlethread your thoughts here?

mynameistechno:
> Do we want the preflight checks on forked projects like this one? If so we should probably tune the configs for what is really relevant here.

I don't think so. The QCOM preflight workflow is for our repos, upstream repos will have their own contributions requirements and checkers. I think we can close this unless someone sees any value in e.g. Semgrep static analysis or something?

mynameistechno:
I should caveat - unless this is a hard fork. But I assume the goal with this fork is to upstream

ricardosalveti:
> I should caveat - unless this is a hard fork. But I assume the goal with this fork is to upstream

Yeah, the goal here is to be as close with upstream as possible.

Will close then.

PR #2, opened 2025-09-25
plat: Add support for Qualcomm Kodiak platform

b49020:
Introduce initial Qualcomm platform support for the Kodiak which is the SoC codename also known by product names SC7280/QCM6490 in upstream.

### DIFF
--- a/core/arch/arm/plat-qcom/conf.mk
+++ b/core/arch/arm/plat-qcom/conf.mk
@@ -1,0 +1,27 @@
+# Qualcomm platform support
+
+PLATFORM_FLAVOR ?= kodiak
+
+$(call force,CFG_GIC,y)
+$(call force,CFG_ARM_GICV3,y)
+$(call force,CFG_SECURE_TIME_SOURCE_CNTPCT,y)
+$(call force,CFG_ARM64_core,y)
+$(call force,CFG_WITH_ARM_TRUSTED_FW,y)
+$(call force,CFG_CORE_ARM64_PA_BITS,40)
+$(call force,CFG_CORE_LARGE_PHYS_ADDR,y)
+$(call force,CFG_CORE_RESERVED_SHM,n)
+$(call force,CFG_QCOM_GENI_UART,y)
+
+ta-targets = ta_arm64
+supported-ta-targets ?= ta_arm64
+
+ifeq ($(PLATFORM_FLAVOR),kodiak)
+include core/arch/arm/cpu/cortex-armv8-0.mk
+$(call force,CFG_TEE_CORE_NB_CORE,8)
+
+CFG_TZDRAM_START ?= 0x1c300000
+CFG_TEE_RAM_VA_SIZE ?= 0x200000
+CFG_TA_RAM_VA_SIZE ?= 0x1c00000
+CFG_TZDRAM_SIZE  ?= (CFG_TEE_RAM_VA_SIZE + CFG_TA_RAM_VA_SIZE)
+CFG_NUM_THREADS  ?= 8
+endif
--- a/core/arch/arm/plat-qcom/main.c
+++ b/core/arch/arm/plat-qcom/main.c
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2025, Linaro Limited
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#include <console.h>
+#include <drivers/gic.h>
+#include <drivers/qcom_geni_uart.h>
+#include <io.h>
+#include <kernel/boot.h>
+#include <platform_config.h>
+
+/*
+ * Register the physical memory area for peripherals etc. Here we are
+ * registering the UART console.
+ */
+register_phys_mem_pgdir(MEM_AREA_IO_NSEC, GENI_UART_REG_BASE,
+			GENI_UART_REG_SIZE);
+
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICD_BASE, GIC_DIST_REG_SIZE);
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICR_BASE, GIC_DIST_REG_SIZE);
+
+register_ddr(DRAM0_BASE, DRAM0_SIZE);
+#ifdef DRAM1_BASE
+register_ddr(DRAM1_BASE, DRAM1_SIZE);
+#endif
+
+static struct qcom_geni_uart_data console_data;
+
+void plat_console_init(void)
+{
+	qcom_geni_uart_init(&console_data, GENI_UART_REG_BASE);
+	register_serial_console(&console_data.chip);
+}
+
+void boot_primary_init_intc(void)
+{
+	gic_init_v3(0, GICD_BASE, GICR_BASE);
+}
+
+void boot_secondary_init_intc(void)
+{
+	gic_init_per_cpu();
+}
--- a/core/drivers/qcom_geni_uart.c
+++ b/core/drivers/qcom_geni_uart.c
@@ -1,0 +1,50 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Qualcomm GENI serial engine UART driver
+ *
+ * Copyright (c) 2025, Linaro Limited
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#include <drivers/qcom_geni_uart.h>
+#include <io.h>
+
+#define GENI_STATUS_REG			0x40
+#define GENI_STATUS_REG_CMD_ACTIVE	BIT(0)
+#define GENI_TX_FIFO_REG		0x700
+#define GENI_TX_TRANS_LEN_REG		0x270
+#define GENI_M_CMD0_REG			0x600
+
+#define GENI_M_CMD_TX			(0x8000000)
+
+static void qcom_geni_uart_putc(struct serial_chip *chip, int ch)
+{
+	struct qcom_geni_uart_data *pd =
+		container_of(chip, struct qcom_geni_uart_data, chip);
+	vaddr_t base = io_pa_or_va(&pd->base, GENI_UART_REG_SIZE);
+	uint64_t timer = timeout_init_us(1000 * 1000);
+
+	while (io_read32(base + GENI_STATUS_REG) & GENI_STATUS_REG_CMD_ACTIVE) {
+		if (timeout_elapsed(timer))
+			break;
+	}
+	io_write32(base + GENI_TX_TRANS_LEN_REG, 1);
+	io_write32(base + GENI_M_CMD0_REG, GENI_M_CMD_TX);
+	io_write32(base + GENI_TX_FIFO_REG, ch);
+}
+
+static const struct serial_ops qcom_geni_uart_ops = {
+	.putc = qcom_geni_uart_putc,
+};
+DECLARE_KEEP_PAGER(qcom_geni_uart_ops);
+
+void qcom_geni_uart_init(struct qcom_geni_uart_data *pd, paddr_t base)
+{
+	pd->base.pa = base;
+	pd->chip.ops = &qcom_geni_uart_ops;
+
+	/*
+	 * Do nothing, debug uart is shared with normal world, everything
+	 * for debug uart initialization is done in the bootloader.
+	 */
+}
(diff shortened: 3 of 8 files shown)


ldts:
thanks @b49020.  could we add an entry in https://github.com/qualcomm-linux/optee_os/blob/qcom-next/.github/workflows/ci.yml?

core/drivers/geni_uart.c
```diff
@@ -0,0 +1,60 @@
...
+{
+	uint64_t timer = timeout_init_us(1000 * 1000);
+	while (io_read32(base + GENI_STATUS_REG) & GENI_STATUS_REG_CMD_ACTIVE) {
+		udelay(10);
+		if (timeout_elapsed(timer))
+			break;
+	}
+}
+
+static void geni_uart_putc(struct serial_chip *chip, int ch)
```
ldts: I'd include wait_txt_done in the putc function instead of adding a static.  (just like it is done https://github.com/OP-TEE/optee_os/blob/master/core/drivers/stm32_uart.c#L68)
is the udelay necessary?

core/include/drivers/geni_uart.h
```diff
@@ -0,0 +1,21 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2024, Linaro Limited
+ */
+#ifndef __DRIVERS_GENI_UART_H
+#define __DRIVERS_GENI_UART_H
+
+#include <types_ext.h>
```
ldts: alphabetical order

core/drivers/qcom_geni_uart.c
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
ldts, line 1: I'd remove the comments and just leave the copyright.

b49020: Let me just keep `Qualcomm GENI serial engine UART driver` comment and drop the other one.

core/drivers/qcom_geni_uart.c
zelvam95, line 1: Many vendors in OPTEE currently have uart driver file name as <vendor_uart.c>. Would it be good to align with them and rename the UART driver for QCOM as qcom_uart.c? Especially given this UART driver can only be used on QCOM Chipsets?

Reference of existing uart driver names in OPTEE:
atmel_uart.c
amlogic_uart.c
mvebu_uart.c
stm32_uart.c
sifive_uart.c

b49020: Sure, I can rename the driver files to `qcom_geni_uart` since I suppose there is another serial driver `qcom_msm_uart` which is needed for older SoCs.

ldts: recently a **vendor** folder appeared in core/drivers (**core/drivers/amd)**. Maybe we could check if this is a one off or the new direction for drivers?

b49020: I would rather opt for all the uart drivers to be put under `core/drivers/serial/` instead since then it will be much more compatible to other projects. We already have driver directories like `core/drivers/crypto/`, `core/drivers/pinctrl/`. I think the `core/drivers/qcom/` will come into picture for drivers that don't share any similarity with other silicon vendors like xpu etc.

I will propose a separate refactoring patch upstream.

core/drivers/geni_uart.c
b49020: Dropped redundant `udelay()` as well.

core/arch/arm/plat-qcom/main.c
```diff
@@ -0,0 +1,47 @@
...
+#include <kernel/panic.h>
+#include <mm/core_mmu.h>
+#include <platform_config.h>
+#include <stdint.h>
+
+/*
+ * Register the physical memory area for peripherals etc. Here we are
+ * registering the UART console.
+ */
+register_phys_mem_pgdir(MEM_AREA_IO_NSEC, CONSOLE_UART_BASE,
```
zelvam95: Should we use register_phys_mem instead of register_phys_mem_pgdir for UART given GENI_UART_REG_SIZE is 16 KB?

core/drivers/qcom_geni_uart.c
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright (c) 2025, Linaro Limited
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#include <console.h>
+#include <drivers/qcom_geni_uart.h>
+#include <io.h>
+#include <keep.h>
+
+#define GENI_STATUS_REG			0x40
```
zelvam95, line 12: Would using tab alignment for all macros (both in .c and .h) make the code more readable?

core/arch/arm/plat-qcom/conf.mk
```diff
@@ -0,0 +1,27 @@
...
+supported-ta-targets ?= ta_arm64
+
+ifeq ($(PLATFORM_FLAVOR),kodiak)
+include core/arch/arm/cpu/cortex-armv8-0.mk
+$(call force,CFG_TEE_CORE_NB_CORE,8)
+
+CFG_TZDRAM_START ?= 0x1c300000
+CFG_TEE_RAM_VA_SIZE ?= 0x200000
+CFG_TA_RAM_VA_SIZE ?= 0x1c00000
+CFG_TZDRAM_SIZE  ?= (CFG_TEE_RAM_VA_SIZE + CFG_TA_RAM_VA_SIZE)
```
zelvam95, line 25: Is this intentional?
CFG_TEE_RAM_VA_SIZE defines the virtual address space size for the TEE core, while CFG_TZDRAM_SIZE defines the total physical secure DRAM allocation.

core/arch/arm/plat-qcom/conf.mk
```diff
@@ -0,0 +1,27 @@
...
+CFG_NUM_THREADS  ?= 8
```
zelvam95, line 26: Would it be better to update like this? (similar to many other vendors)
CFG_NUM_THREADS ?= $(CFG_TEE_CORE_NB_CORE)

core/drivers/qcom_geni_uart.c
zelvam95, line 1: Reg. UART Driver -> Are we assuming the QUP FW is loaded already for UART & Clock/TLMM Init is done by TF-A/BOOTLOADER for UART? Would it be good to include that assumption as a comment somewhere?

core/arch/arm/plat-qcom/main.c
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#include <console.h>
+#include <drivers/gic.h>
+#include <drivers/qcom_geni_uart.h>
+#include <io.h>
+#include <kernel/boot.h>
+#include <kernel/delay.h>
+#include <kernel/panic.h>
```
zelvam95: Have we included some header files by default even if we don't "require" them currently for future potential usage?

For instance panic.h might not be required if we're not calling "panic()"?

core/arch/arm/plat-qcom/main.c
```diff
@@ -0,0 +1,47 @@
...
+#include <stdint.h>
+
+/*
+ * Register the physical memory area for peripherals etc. Here we are
+ * registering the UART console.
+ */
+register_phys_mem_pgdir(MEM_AREA_IO_NSEC, CONSOLE_UART_BASE,
+			GENI_UART_REG_SIZE);
+
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICD_BASE, GIC_DIST_REG_SIZE);
```
zelvam95, line 21: If GICD and GICR would be contiguous without any gap inbetween them (I assume its usually that way in GICv3?), would it be better to have a single unified mapping here using register_phys_mem_pgdir -> GIC_BASE to GIC_SIZE

core/arch/arm/plat-qcom/main.c
```diff
@@ -0,0 +1,47 @@
...
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICR_BASE, 0x100000);
+
+register_ddr(DRAM0_BASE, DRAM0_SIZE);
+register_ddr(DRAM1_BASE, DRAM1_SIZE);
```
zelvam95, line 26: Would all chipsets have split DDR like this? Would it be better if we can check if DRAM*_BASE is defined and then include it?

core/arch/arm/plat-qcom/main.c
b49020: It is rather better to use an API which takes care of the page alignment itself.

core/drivers/qcom_geni_uart.c
b49020: It's the general practice followed by OP-TEE project to use tabs here.

core/arch/arm/plat-qcom/conf.mk
b49020: Yeah, here I wanted to specifically map TA RAM to the second pseudo IMEM window which is after the first 5MB first window.

core/arch/arm/plat-qcom/conf.mk
b49020: I intentionally kept it separate to avoid the notion that the number of threads is equivalent to number of cores which isn't the case. Depending upon memory pressure in OP-TEE then you can always lower/increase the number of threads.

core/drivers/qcom_geni_uart.c
b49020: Have a look at the boot flow here: https://github.com/b49020/trusted-firmware-a/blob/qti/rb3gen2/docs/plat/qti/rb3gen2.rst. We are reusing XBL stage which is responsible for loading QUP FW and initializing clock/TLMM. I can add the comment in the driver.

core/arch/arm/plat-qcom/main.c
b49020: Okay, I will drop this redundant header include here.

core/arch/arm/plat-qcom/main.c
b49020: I don't think so we need to make any assumption here about the address map.

core/arch/arm/plat-qcom/main.c
b49020: That's reasonable, I can do it for the `DRAM1_BASE` since we expect atleast `DRAM0_BASE` to be defined.

core/drivers/qcom_geni_uart.c
b49020: Done

core/arch/arm/plat-qcom/main.c
zelvam95: Would it be better if we just use CONSOLE_UART_SIZE here? instead of "GENI_UART_REG_SIZE"? & currently base and size are coming from 2 different files. Is it intentional?

core/arch/arm/plat-qcom/main.c
zelvam95: Sure, Got it. GICD_BASE & GICR_BASE both are using size as GIC_DIST_REG_SIZE -> Is this a safe assumption?

core/drivers/qcom_geni_uart.c
```diff
@@ -0,0 +1,50 @@
...
+	io_write32(base + GENI_M_CMD0_REG, GENI_M_CMD_TX);
+	io_write32(base + GENI_TX_FIFO_REG, ch);
+}
+
+static const struct serial_ops qcom_geni_uart_ops = {
+	.putc = qcom_geni_uart_putc,
+};
+DECLARE_KEEP_PAGER(qcom_geni_uart_ops);
+
+void qcom_geni_uart_init(struct qcom_geni_uart_data *pd, vaddr_t base)
```
zelvam95: The second param should ideally be paddr_t base since thats the "pa"?

core/arch/arm/plat-qcom/main.c
b49020: Yeah that true since it's a common practice across platforms.

core/arch/arm/plat-qcom/main.c
b49020: > Would it be better if we just use CONSOLE_UART_SIZE here? instead of "GENI_UART_REG_SIZE"? 

I would rather rename base as `GENI_UART_REG_BASE`.

> currently base and size are coming from 2 different files. Is it intentional?

It's due to current driver API design which doesn't allow size to be passed as part of API. And the normal convention is that the driver code should be independent of platform code.

b49020:
Thanks for all the feedback, further review can happen upstream here: https://github.com/OP-TEE/optee_os/pull/7547. Let me merge this PR for staging tree as of now.

PR #7, opened 2026-04-04
plat-qcom: Add diagnostic logging support

zelvam95:
This patch series introduces a diagnostic ring buffer driver for Qualcomm
platforms, enabling persistent early boot logging that survives warm resets
and facilitates post-mortem analysis of system failures.

The implementation is structured as two patches:
1. Core driver with self-contained memory management
2. Platform enablement for Kodiak and Lemans with optimized memory layout

Key capabilities:
- Persistent circular buffer in IMEM survives warm resets
- Automatic overflow handling with wrap counter
- Download mode detection preserves crash diagnostics
- Self-contained driver with automatic memory registration
- Seamless OP-TEE trace subsystem integration

Memory layout:
- Kodiak: Buffer at 0x14696000 (within 100KB IMEM)
- Lemans: Buffer at 0x146AF000 (within 200KB IMEM)
- Common: Metadata at 0x14680720, TCSR at 0x1FD3000

This implementation enables comprehensive system debugging by providing
persistent logs accessible to bootloaders and diagnostic tools across
system resets.


b49020:
> Memory layout:
> 
> * Kodiak: Buffer at 0x14696000 (within 100KB IMEM)
> * Lemans: Buffer at 0x146AF000 (within 200KB IMEM)
> * Common: Metadata at 0x14680720, TCSR at 0x1FD3000

@zelvam95 how is this allocation done on IMEM for the diag buffer? `0x14680000` is actually the load address for TF-A BL2/ U-Boot SPL on Lemans platform as TZ image. I believe there is overlap going on here.

zelvam95:
> > Memory layout:
> > 
> > * Kodiak: Buffer at 0x14696000 (within 100KB IMEM)
> > * Lemans: Buffer at 0x146AF000 (within 200KB IMEM)
> > * Common: Metadata at 0x14680720, TCSR at 0x1FD3000
> 
> @zelvam95 how is this allocation done on IMEM for the diag buffer? `0x14680000` is actually the load address for TF-A BL2/ U-Boot SPL on Lemans platform as TZ image. I believe there is overlap going on here.

Hi @b49020,

I referred to QTEE Code & in QTEE, this region is being used for DIAG Buffer. Its reserved/marked down for that purpose. Have we changed the layout/do we have a separate layout for our current open source boot loader stack?

b49020:
> > > Memory layout:
> > > 
> > > * Kodiak: Buffer at 0x14696000 (within 100KB IMEM)
> > > * Lemans: Buffer at 0x146AF000 (within 200KB IMEM)
> > > * Common: Metadata at 0x14680720, TCSR at 0x1FD3000
> > 
> > 
> > @zelvam95 how is this allocation done on IMEM for the diag buffer? `0x14680000` is actually the load address for TF-A BL2/ U-Boot SPL on Lemans platform as TZ image. I believe there is overlap going on here.
> 
> Hi @b49020,
> 
> I referred to QTEE Code & in QTEE, this region is being used for DIAG Buffer. Its reserved/marked down for that purpose. Have we changed the layout/do we have a separate layout for our current open source boot loader stack?

This DIAG buffer is internal to QTEE memory layout where it is put on IMEM I think. How does the crashdump flow gets to know this address? Is it hard coded in the tooling somewhere for each SoC?

zelvam95:
> > > > Memory layout:
> > > > 
> > > > * Kodiak: Buffer at 0x14696000 (within 100KB IMEM)
> > > > * Lemans: Buffer at 0x146AF000 (within 200KB IMEM)
> > > > * Common: Metadata at 0x14680720, TCSR at 0x1FD3000
> > > 
> > > 
> > > @zelvam95 how is this allocation done on IMEM for the diag buffer? `0x14680000` is actually the load address for TF-A BL2/ U-Boot SPL on Lemans platform as TZ image. I believe there is overlap going on here.
> > 
> > 
> > Hi @b49020,
> > I referred to QTEE Code & in QTEE, this region is being used for DIAG Buffer. Its reserved/marked down for that purpose. Have we changed the layout/do we have a separate layout for our current open source boot loader stack?
> 
> This DIAG buffer is internal to QTEE memory layout where it is put on IMEM I think. How does the crashdump flow gets to know this address? Is it hard coded in the tooling somewhere for each SoC?

Its not hard coded; Its rather retrieved from another IMEM address. For QTEE DIAG Log, they write the DIAG Buffer start address at offset 0x720 & crashscope/tools read that offset, understand where the DIAG Log is present and then does the parsing.

b49020:
> > > > > Memory layout:
> > > > > 
> > > > > * Kodiak: Buffer at 0x14696000 (within 100KB IMEM)
> > > > > * Lemans: Buffer at 0x146AF000 (within 200KB IMEM)
> > > > > * Common: Metadata at 0x14680720, TCSR at 0x1FD3000
> > > > 
> > > > 
> > > > @zelvam95 how is this allocation done on IMEM for the diag buffer? `0x14680000` is actually the load address for TF-A BL2/ U-Boot SPL on Lemans platform as TZ image. I believe there is overlap going on here.
> > > 
> > > 
> > > Hi @b49020,
> > > I referred to QTEE Code & in QTEE, this region is being used for DIAG Buffer. Its reserved/marked down for that purpose. Have we changed the layout/do we have a separate layout for our current open source boot loader stack?
> > 
> > 
> > This DIAG buffer is internal to QTEE memory layout where it is put on IMEM I think. How does the crashdump flow gets to know this address? Is it hard coded in the tooling somewhere for each SoC?
> 
> Its not hard coded; Its rather retrieved from another IMEM address. For QTEE DIAG Log, they write the DIAG Buffer start address at offset 0x720 & crashscope/tools read that offset, understand where the DIAG Log is present and then does the parsing.

Does that mean whole of IMEM is reserved for the crashdump tooling? Isn't XBL using any of the IMEM regions too? I think the memory layout has to be studied further. Would you be able to put the memory layout on a confluence page such that we can have further alignment?

zelvam95:
**Kodiak:**
 
100K starting 0x14680000

Start Address    Page Size       Region
14680000         24K             Monitor Code/Data
14686000         20K             Reserved
1468B000         40K - 64 bytes  QSEE Code/Data
14694FBC         0 bytes         Dummy cookie to check overlap. New cookies will be added below this cookie.
14694FC0         4 bytes         Device Storage Type
14694FC4         48 bytes        Measured Boot Info
14694FF4         4 bytes         Sec Mem Info
14694FF8         8 bytes         STACK_CANARY_INFO
14695000         4K              OEM_IMEM
14696000         12K             TZ Diag

**Lemans:**

200K starting 0x14680000

Start Address    Page Size       Region
14680000         24K             Monitor Code/Data
14686000         36K             Reserved
146A8000         24K - 20 bytes  QSEE Code/Data
146ADFEC         4 bytes         Dummy cookie to check overlay
146ADFF0         4 bytes         Device Storage Type
146ADFF4         4 bytes         Sec Mem Info
146ADFF8         8 bytes         STACK_CANARY_INFO
146AE000         4K              OEM_IMEM
146AF000         12K             TZ Diag

zelvam95:
Hi @b49020,

This info is documented and available internally already. I have shared high level split above. The OCIMEM is split across different sub systems for different use-cases as required. We can review this and align on this.

b49020:
@zelvam95 thanks for gathering that info, I have updated TF-A BL2 for Lemans to load and execute from pIMEM instead here: https://github.com/qualcomm-linux/trusted-firmware-a/pull/15. Feel free to test your changes with that.

b49020:
Also, TF-A BL2 on Kodiak was already executing from pIMEM as you can see here: https://github.com/qualcomm-linux/trusted-firmware-a/blob/qcom-next/plat/qti/hoya/kodiak/rb3gen2/inc/platform_def.h#L15

core/drivers/qcom_diag_log.c
```diff
@@ -0,0 +1,164 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
ldts: can you move the driver to core/drivers/qcom/

core/drivers/qcom_diag_log.c
```diff
@@ -0,0 +1,164 @@
...
+	tcsr_reg = (uint32_t *)io_pa_or_va(&tcsr_pa_va, sizeof(uint32_t));
+	if (!tcsr_reg)
+		return false;
+
+	reg_value = io_read32((vaddr_t)tcsr_reg);
+
+	return (reg_value == DLOAD_MAGIC_COOKIE);
+}
+
+void qcom_diag_log_init(void)
```
ldts: should this function return an error code?

core/arch/arm/plat-qcom/main.c
```diff
@@ -32,8 +35,19 @@ register_ddr(DRAM1_BASE, DRAM1_SIZE);
...
+#ifdef CFG_QCOM_DIAG_LOG
+void plat_trace_ext_puts(const char *str)
+{
+	qcom_diag_log_puts(str);
+}
+#endif
+
 void plat_console_init(void)
 {
+#ifdef CFG_QCOM_DIAG_LOG
```
ldts: I think this makes the code easier to read - the compiler will optimize it out
```
static void plat_diag_log_init()
{
#ifdef CFG_QCOM_DIAG_LOG
	int err = qcom_diag_log_init();
    
     if (err)
         EMSG("Diag log not available");
#endif
}

void plat_console_init(void)
{
	plat_diag_log_init()
	qcom_geni_uart_init(&console_data, GENI_UART_REG_BASE);
	register_serial_console(&console_data.chip);
}
```

core/arch/arm/plat-qcom/main.c
```diff
@@ -32,8 +35,19 @@ register_ddr(DRAM1_BASE, DRAM1_SIZE);
 
 static struct qcom_geni_uart_data console_data;
 
+#ifdef CFG_QCOM_DIAG_LOG
+void plat_trace_ext_puts(const char *str)
```
ldts, line 33: `plat_write_log()` instead?

core/arch/arm/plat-qcom/platform_config.h
```diff
@@ -22,9 +22,19 @@
 
 #define RAMBLUR_PIMEM_REG_BASE		UL(0x610000)
 #define SEC_PRNG_REG_BASE		UL(0x010D1000)
+
+#define IMEM_BASE			UL(0x14680000)
+#define IMEM_DIAG_OFFSET		UL(0x720)
+#define DIAG_LOG_START_INFO		(IMEM_BASE + IMEM_DIAG_OFFSET)
```
ldts: not sure if this should go here. offsets typically go in the driver and so the driver could reference IMEM_BASE + offset.

TCSR_BOOT_MISC_DETECT ...is this an offset on top of some TCSR region?

core/drivers/qcom_diag_log.c
```diff
@@ -0,0 +1,164 @@
...
+		io_write32((vaddr_t)&diag_info_addr[1], DIAG_SIZE);
+	}
+
+	dsb();
+}
+
+/*
+ * Write a string to the diagnostic log buffer
+ */
+void qcom_diag_log_puts(const char *str)
```
ldts: I assume this can be called from multiple contexts? 

`	mutex_lock(&diag->lock);
`

core/drivers/qcom_diag_log.c
```diff
@@ -0,0 +1,164 @@
...
+	if (!str)
+		return;
+
+	diag = get_diag_region();
+	if (!diag)
+		return;
+
+	if (diag->magic != DIAG_MAGIC_INIT)
+		return;
+
```
ldts:
```
for (p = str; *p; p++) {
    diag->log_buf[diag->offset++] = *p;
    if (diag->offset >= diag->ring_len) {
        diag->offset = 0;
        if (diag->wrap < UINT32_MAX)
            diag->wrap++;
    }
}
```

core/include/drivers/qcom_diag_log.h
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#ifndef QCOM_DIAG_LOG_H
+#define QCOM_DIAG_LOG_H
+
+#include <tee_api_types.h>
+#include <types_ext.h>
+
+/* Initialize the diagnostic logging system */
```
ldts: no need for these comments

core/drivers/qcom_diag_log.c
```diff
@@ -0,0 +1,164 @@
...
+			SMALL_PAGE_SIZE);
+
+#define DIAG_VERSION		1          /* Structure version */
+#define DIAG_MAGIC_INIT		0xBEEF     /* Buffer initialized and ready */
+#define DIAG_MAGIC_FAILED	0xDEAD     /* Initialization failed */
+#define DIAG_MAGIC_DLOAD	0xCAFE     /* DLOAD mode - logging disabled */
+#define DLOAD_MAGIC_COOKIE	0x10       /* Download mode detection value */
+
+/*
+ * struct diag - Main diagnostic region structure
```
ldts: no need for these comments.  IMO we should use standard semantics for the circular buffer instead of these definitions.

Would you mind using the usual head, tail? maybe check how linux implements a simple one (include/linux/circ_buf.h)
```

struct diag {
	uint32_t magic;
	uint32_t ring_len;    /* must be power of 2 */
	uint32_t head;        /* write index, monotonically increasing */
	uint32_t tail;        /* read index,  monotonically increasing */
	uint32_t overrun;
	struct mutex lock;
	uint8_t  log_buf[];
};
```
could we also provide a reader or it makes no sense?

Otherwise, if we are not going to  provide a reader lets not call them overruns and we dont need the tail...they are just wraps I think?

```	mutex_lock(&diag->lock);

	for (p = str; *p; p++) {
		head = diag->head++;
		if (head && !(head & (diag->ring_len - 1)))
			diag->wrap++;
		diag->log_buf[head & (diag->ring_len - 1)] = *p;
	}

	dsb();
	mutex_unlock(&diag->lock);

```

core/drivers/qcom_diag_log.c
zelvam95: This is hooked in plat_trace_ext_puts available in OP-TEE where we already have the required mutex locks.

core/arch/arm/plat-qcom/main.c
zelvam95: This is not a function I am introducing. plat_trace_ext_puts is a platform hook called as part of console framework & I am plugging in our diag log driver putc over there so it gets called each time we do EMSG/DMSG/IMSG/etc.

core/arch/arm/plat-qcom/main.c
zelvam95: Sure, will update it this way.

core/drivers/qcom_diag_log.c
zelvam95: Will move it.

core/drivers/qcom_diag_log.c
zelvam95: Reader not required. version was suggested during wider discussion with SSG Team.

This log driver would be used across arch + there will be tooling built around this. We don't need mutex around this. Mutex is there in the calling place for diag_log_putc (plat_trace_ext_puts). Similarly tail is not required.

ring_offset & ring_len also have a purpose & it will be used by HLOS/tooling infra built around this. We can potentially introduce more variables in the future version before log_buf & ring_offset would print to exact start of log_buf.

Also, ring_len gives out the log buf size which can be used by tooling infra.

This struct was aligned on based on discussion with SSG Team & we plan to re-use same infra/structure across TF-A & OP-TEE. I got a comment to make both version & magic as uint32_t which I will be updating.

core/drivers/qcom_diag_log.c
zelvam95: Even if we return error code, all we can do is log it in calling place... I think since we're already logging here during successful init + magic is updated to indicate any initialization failure which can be used by tooling/HLOS drivers as required, so I think we needn't necessarily return an error here.

Let me know if you feel its better to return for any reason? & we can plan accordingly;

core/drivers/qcom_diag_log.c
zelvam95: Will update this;

core/include/drivers/qcom_diag_log.h
zelvam95: Sure, will clean up all comments;

core/arch/arm/plat-qcom/platform_config.h
zelvam95: The IMEM_BASE/IMEM_OFFSET/TCSR would change for each platforms.
Maybe I will have the IMEM_BASE here since thats generic can be used in other dirvers.
Everything else I will move to driver specific platform/target header file.

core/drivers/qcom_diag_log.c
ldts: I dont understand your points

1) the diag region is global memory. every write access to it must be protected by a mutex. the mutex must belong to the structure.
2) the identifiers/names in the structure seem to come from a school primer on computer science. we can/should do better. if there is no reader this is not really a ring buffer, it is just a circular buffer (at least for me).

These sort of definitions seem more appropriate to me (layout may vary)
```
 struct circ_wo_buf {                                                                                                                                                                                                    
      uint8_t     *buf;                                                                                                                                                                                                
      size_t       size;       
      unsigned int head;       
      unsigned int wrap;       
      struct mutex lock;       
  };        
```

```
struct diag_buf {
      uint16_t version;
      uint 16_t magic
      struct circ_wo_buf wo_cbuf;  /*maybe also *wo_cbuf in case address varies with version */
}
```

zelvam95: Hi Jorge,

**Reg. variable naming/structure definition:**

- I was clarifying on the layout as to why its required and about the tooling requirements downstream for retaining a particular layout. 
- Also, the current struct variables like offset, len etc. namings are already used in BCM's implementation which is available upstream. Its a local variable there & I have made it as part of the header, thats the main diff between QCOM vs BCM's impl.. 
- I feel different persons might have different perspective/preferences in this. Do you feel the BCM's impl is not upto the mark and their naming is also like from a school primer? I used that as reference since thats the only implementation available upstream which uses this plat_trace_ext_puts.
- **Anyways, I am open to make minor modifications as suggested above with nested structure + slight name changes.**. 

**Reg. mutex - Your concern about global memory access is valid in general, but in this specific case:**
- __The trace infrastructure already provides serialization__: All calls to `qcom_diag_log_puts()` go through `trace_ext_puts()` which ensures serialized access.
- __Single controlled entry point__: Unlike general-purpose drivers, this has a single entry point through the trace infrastructure, which acts as the serialization gatekeeper.
- __This follows the established pattern in OP-TEE__:
- - BCM's elog implementation (`plat-bcm/bcm_elog.c`) uses the same approach: global memory buffer, no mutex, relies on trace infrastructure

**Note:** Would request you to please check the trace_ext_puts function -> where plat_trace_ext_puts is called & also check the BCM's implementation of log driver where a similar diag log impl is already available and upstreamed.

zelvam95:
```
struct circ_wo_buf {
	unsigned int buf_off;
	size_t size;
	unsigned int head;
	unsigned int wrap;
	uint8_t buf[];
};
```

```
struct diag_buf {
	uint32_t version;
	uint32_t magic;
	struct circ_wo_buf wo_cbuf;
};
```

zelvam95: Planning to update above way now. We're using "flexible array" at the end & so can't update as uint8_t* buf at the start. buf_off is to give the "offset" to buf[]. In future, even if we add new fields, the external toolings etc would read the buf_off to identify buf[] start and so wouldn't have issues. 

**Let me know what you think of this struct + mutex requirements.**

ldts: WRT to the lock, yes, drop it from the circular buffer definition - I didnt notice we were using the trace infrastructure (I should have clone the repo and read it)

WRT to the names, yes, **do use head for semantics and define a circular buffer explicitly.**
These are not slight name changes - these are software semantics that make the code readable - the distance between the word "offset" or "head" is similar to the difference between an atom and a horse.


I am not part of the SSG discussions on the structure defintions, but do not use a flat layout - use abstractions...ie see below for another example....

```
struct diag_hdr {
   uint16_t version;
   uint16_t magic;
};

struct circ_buf_conf {
   uint32_t base;
   uint32_t size;
};

struct circ_wo_buf {
       uint32_t wrap;
       uint32_t head;
       uint8_t buf[];
};

struct diag {
       struct diag_hdr hdr;
       struct circ_buf_conf conf;
       struct circ_wo_buf wo_cbuf;
};
```

zelvam95: Yes, I am using trace infra. Mutex part is clear now :).

**Reg. structure variable naming:**
I used "BCMs" implementation of the log driver as reference for this & in their implementation if you check, you'd notice they are using "offset" instead of "head" for a very similar log driver that they have. I assumed what's already upstream available can be used as a standard for reference but probably not all changes upstream follow these software semantics. I mentioned them as slight name changes because otherwise if it was considered as part of sw semantics, then they should have knocked down the BCMs impl when it was posted for review?

**Note -** I am just trying to highlight the point of reference - that it was already avail upstream.  

Anyways, You can check the latest patch where I have modified to using head instead of offset. I will incrementally do changes as suggested above with abstractions & post same for review.

ldts: yeah, don't rely on what's upstream :) as you know (better than I do!) we have many platforms to upstream so the better our abstractions (the more indirections) the easier will be to extend and maintain.

core/drivers/qcom/diag/diag_log.c
```diff
@@ -0,0 +1,155 @@
...
+	diag = get_diag_region();
+	if (!diag)
+		return;
+
+	if (diag->magic != DIAG_MAGIC_INIT)
+		return;
+
+	for (p = str; *p; p++) {
+		diag->wo_cbuf.buf[diag->wo_cbuf.head++] = *p;
+		if (diag->wo_cbuf.head >= diag->wo_cbuf.size) {
```
ldts: it is safer to mask rather than this sort of checks to avoid corrupting memory on glitches.

we can use this sort of construct as long as we can guarantee the size of the buffer is ^2 I think. then there is zero risk or running over the memory

```
	for (p = str; *p; p++) {
		head = diag->head++;
		if (head && !(head & (diag->wo_cbuf.size - 1)))
			diag->wrap++;
		diag->log_buf[head & (diag->wo_cbuf.size - 1)] = *p;
	}
```

ldts:
Please remove the driver initialization from main.c - just initialize it in the driver itself (driver_init, early_init or whatever is appropiate). 
main should just implement plat_trace_ext_puts. 

then qcom_diag_log_puts should keep a static pointer to a valid diag. And we shouldnt need to check for the magic every time...

core/drivers/qcom_diag_log.c
ldts: yes, at least do generate a WARN on diag being configured but not available. 
Tracing errors is important on secure platforms.

core/drivers/qcom/diag/diag_log.c
zelvam95: Unfortunately, we cannot guarantee power-of-2 as size for these buffers. There are many existing platforms where this buffer size is not power of 2. Also, We have some header which we have to reduce from DIAG SIZE. So, lets say even if DIAG Size overall is 8KB (power of 2). If we subtract the header. Then the actual circ buffer size would become 4KB.

In our current case, the DIAG size is 12KB
 __Wasted space__: Rounds down to power-of-2 (e.g., 12KB → 8KB usable)
 __~33% loss__: For 12KB DIAG_SIZE, loses 4KB

For 8KB diag log we'd lose 50% and so on. 

This region is ideally expected to be access protected (write access is restricted to AP-S). So, would it be safe to assume SEL1/SEL3 will not corrupt this? Or at least the risk is comparatively low? (Though ideally it is possible in some cases?)

zelvam95:
> Please remove the driver initialization from main.c - just initialize it in the driver itself (driver_init, early_init or whatever is appropiate). main should just implement plat_trace_ext_puts.

I would ideally prefer if this diag log is available as early as possible & I assumed that it'd be good to have this at same point of console init. Even BCM's implementation which is upstream, follows exact same approach where its init is done from main.c.

In PRODUCTION -> Console logs would be disabled & DIAG Logs could be our only way for debug & ideally it'd be better to have this enabled as early as possible to capture everything?

> then qcom_diag_log_puts should keep a static pointer to a valid diag. And we shouldnt need to check for the magic every time...

This part of the comment I will check further & try to handle this.

ldts:
> > Please remove the driver initialization from main.c - just initialize it in the driver itself (driver_init, early_init or whatever is appropiate). main should just implement plat_trace_ext_puts.
> 
> I would ideally prefer if this diag log is available as early as possible & I assumed that it'd be good to have this at same point of console init. Even BCM's implementation which is upstream, follows exact same approach where its init is done from main.c.
> 
> In PRODUCTION -> Console logs would be disabled & DIAG Logs could be our only way for debug & ideally it'd be better to have this enabled as early as possible to capture everything?
> 
> > then qcom_diag_log_puts should keep a static pointer to a valid diag. And we shouldnt need to check for the magic every time...
> 
> This part of the comment I will check further & try to handle this.

so why a driver? why not just a log feature...drivers (except the console) typically have the syscalls to decide when to load.

@b49020 also do we need to integrate the tracing feature with the FFA console? I dont see any hooks there...do you know why?

b49020:
> @b49020 also do we need to integrate the tracing feature with the FFA console?

There isn't anything like FF-A console since there isn't any ABI requirement with TF-A there. But as you said for this diag feature, let's add support in the common platform code rather than a driver since it's not really running in a driver context but all the OP-TEE contexts.

ldts:
> > @b49020 also do we need to integrate the tracing feature with the FFA console?
> 
> There isn't anything like FF-A console since there isn't any ABI requirement with TF-A there. But as you said for this diag feature, let's add support in the common platform code rather than a driver since it's not really running in a driver context but all the OP-TEE contexts.

there is exactly an FF-A console (as an alternative to the console or the semihosting console)
https://github.com/OP-TEE/optee_os/commit/d4a87690eba3c5031f5ac9c7592bb7d2d150697d

image

none provide the hooks to trace.

core/drivers/qcom/diag/diag_log.c
ldts: um, ok...it is a shame that the buffers cant be a power of 2.

b49020:
> there is exactly an FF-A console (as an alternative to the console or the semihosting console)

Ah I see, that's news to me. Not sure how that FF-A console manages sync among different SPs as each can clobber other console logs. Maybe there is SP identifier.

Anyhow, let's not tie this with FF-A since this feature is expected to work on platforms which don't support FF-A too. BTW, this is related to Qualcomm crashdump tooling as well, so we need to support diagnostic logs that the tooling understands during a crash.

Also, thinking about it more the driver vs platform code. I think similar to UART drivers, let's keep the diag as a special driver code.

core/drivers/qcom/diag/kodiak/diag_target.h
```diff
@@ -0,0 +1,18 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#ifndef DIAG_TARGET_H
+#define DIAG_TARGET_H
+
+#define IMEM_BASE		UL(0x14680000)
```
b49020: Let's move this macro to platform specific config file instead.

core/drivers/qcom/diag/kodiak/diag_target.h
```diff
@@ -0,0 +1,18 @@
...
+#define IMEM_DIAG_OFFSET	UL(0x720)
+#define DIAG_LOG_START_INFO	(IMEM_BASE + IMEM_DIAG_OFFSET)
+
+#define TCSR_BOOT_MISC_DETECT	UL(0x1FD3000)
+
+#define DIAG_BASE		UL(0x14696000)
```
b49020: Let's move this macro to platform specific config file too. After that all other macros can be moved to common header and avoid any duplication.

ldts:
> Also, thinking about it more the driver vs platform code. I think similar to UART drivers, let's keep the diag as a special driver code.

I dont think it should be a driver...it is just writing raw to a qcom specific memory region (no protocols, no registers, no locking, nothing...just init it early and it is good to go.)

zelvam95:
> > Also, thinking about it more the driver vs platform code. I think similar to UART drivers, let's keep the diag as a special driver code.
> 
> I dont think it should be a driver...it is just writing raw to a qcom specific memory region (no protocols, no registers, no locking, nothing...just init it early and it is good to go.)

I originally thought its similar to UART driver, and kept it here; But as Jorge pointed out, it doesn't really have any registers + no locking/etc. as well & the current logging drivers avail in upstream also has put it under plat-*/ and not drivers/... So I think moving it there makes sense & planning to do that change.

@b49020, Hope thats fine. If incase there are any stronger reasons to put it under driver, please do let us know so we can plan accordingly;

ldts:
just for future reference: https://github.com/OP-TEE/optee_os/issues/7775

core/drivers/qcom/diag/kodiak/diag_target.h
zelvam95: I have updated DIAG Base update based on IMEM BASE + SIZE + DIAG SIZE; Its usually kept at end of IMEM uniformly on all chipsets for QTEE & Tried to re-use that here; Hope its fine, let me know if any comments on current approach and will fix it incrementally.

core/drivers/qcom/diag/kodiak/diag_target.h
zelvam95: Done

core/drivers/qcom_diag_log.c
zelvam95: Have added WARNs/EMSGs in the diag driver & have updated to using IS_ENABLED checks and removed config checks. Please check latest patch and let me know if any comments;

core/arch/arm/plat-qcom/diag_log.c
```diff
@@ -0,0 +1,135 @@
...
+	for (p = str; *p; p++) {
+		diag->wo_cbuf.buf[diag->wo_cbuf.head++] = *p;
+		if (diag->wo_cbuf.head >= diag->conf.buf_size) {
+			diag->wo_cbuf.head = 0;
+			if (diag->wo_cbuf.wrap < UINT32_MAX)
+				diag->wo_cbuf.wrap++;
+		}
+	}
+
+	dsb();
```
b49020: Not sure why we need a barrier for every diag log, it can be quite expensive one. The logs would be useful after a crash only when you set DIAG_MAGIC_DLOAD and I see the barrier already there.

b49020:
> @b49020, Hope thats fine. If incase there are any stronger reasons to put it under driver, please do let us know so we can plan accordingly;

That's fine with me.

core/arch/arm/plat-qcom/diag_log.c
```diff
@@ -0,0 +1,135 @@
...
+	diag = get_diag_region();
+	if (!diag)
+		return;
+
+	memset(diag, 0, DIAG_SIZE);
+
+	diag->hdr.version = DIAG_VERSION_V1;
+	diag->hdr.magic = DIAG_MAGIC_INIT;
+	diag->conf.buf_offset = offsetof(struct diag, wo_cbuf.buf);
+	diag->conf.buf_size = (DIAG_SIZE - diag->conf.buf_offset) & (~0xF);
```
ldts: I think this is more clear - what do you think?

`	diag->conf.buf_size = ROUNDDOWN2(DIAG_SIZE - diag->conf.buf_offset, 16);`

ldts:
thanks @zelvam95 , this looks really good now.
after you test it again and check the barrrier issue that @b49020 mentioned, you can add my

`Sign-off-by: Jorge Ramirez-Ortiz <jorge.ramirez@oss.qualcomm.com>`

maybe rename the commit (since this is not a driver anymore)

ldts:
@zelvam95  maybe you want to pick this up to your series and use the new hook instead - see https://github.com/OP-TEE/optee_os/pull/7783

core/arch/arm/plat-qcom/diag_log.c
zelvam95: Memory barrier ensures log writes are visible to other observers
(e.g., P-TA reading logs) before returning. While this adds overhead
to each log operation, it's necessary for coherency when logs are
accessed concurrently by other components.

We have one dsb at end of every log puts call and I believe its acceptable given above? Can I leave this as it is? Let me know your suggestion.

zelvam95: Also, the DLOAD Magic check is useful for certain targets only, while for certain targets, the dumps are collected from XBL Stage itself (and so it doesn't come to OP-TEE) itself in DLOAD Path... (Before dump collection). Considering these, I believe its Ok to have the dsb for now and clean up later if/as required.

core/arch/arm/plat-qcom/diag_log.c
zelvam95: Agree, will update with this.

zelvam95:
> @zelvam95 maybe you want to pick this up to your series and use the new hook instead - see [OP-TEE/optee_os#7783](https://github.com/OP-TEE/optee_os/pull/7783)

Have included this & rebased on top of this :).

zelvam95:
> thanks @zelvam95 , this looks really good now. after you test it again and check the barrrier issue that @b49020 mentioned, you can add my
> 
> `Sign-off-by: Jorge Ramirez-Ortiz <jorge.ramirez@oss.qualcomm.com>`
> 
> maybe rename the commit (since this is not a driver anymore)

Thanks Jorge. Have addressed all comments and have included signed off by. dsb() comment + HLOS P-TA changes in-progress. We can plan to merge post that.

core/arch/arm/plat-qcom/diag_log.c
b49020: I still think its rather better to have a memory barrier from reader point of view. Also, since it's a debugging feature it is preferred to not have much logging overhead since sometimes issues under debug may be timing critical as well.

The reboot cycle in case of crashdump in itself should take care of the writes reaching out to the destination buffers.

zelvam95: Done, removed it.

zelvam95:
Update - The plan now is to merge this DIAG Driver to staging tree. Crashdump tooling testing would be done as part of Crashdump PR.

HLOS P-TA plan has been dropped for timebeing and it can be explored/enabled later as required.

b49020:
@zelvam95 there is a merge conflict, please rebase.

zelvam95:
> @zelvam95 there is a merge conflict, please rebase.

Yes, it was due to the Lemans Fuseprov change we merged recently. I have rebased this PR now & have again validated boot up test on Lemans-EVK.

b49020 (approved):

PR #8, opened 2026-04-12
drivers: hwkm: initial draft for hwkm HAL

qc-azarrabi:
This is just for discussion.


ldts:
@qc-azarrabi I know it is a pain but typedefs are not welcome. Also ident with TABs. You can check all these things by

```
cd optee_os
scripts/checkpatch.sh HEAD
```

could you add a use case interface? ie the huk generation? or some crypto interface?

qc-azarrabi:
@ldts Yes definitely. This needs extensive cleanup and re-write. I'll update in a clean RP.

PR #10, opened 2026-04-25
drivers/crypto: add Qualcomm Hardware Key Manager (HWKM) driver

qc-azarrabi:
This patch adds a new driver for the Qualcomm Hardware Key Manager (HWKM), a hardware IP block found on Qualcomm SoCs that provides hardware-enforced key storage and cryptographic key derivation.

Background
-------------
HWKM maintains a key table whose slots are never accessible in plaintext to software above the security level at which a key was provisioned. Each slot carries a hardware-enforced policy word that controls which operations, algorithms, and security domains may use it. The communicates with the hardware is through a command/response FIFO interface.

What this patch does
-----------------------
The driver implements the full HWKM command set:
  - NIST_KEYGEN - generate a fresh key into a slot via the HW PRNG.
  - SYSTEM_KDF - derive a child key from a KDK slot using the hardware KDF, with optional BSVE binding.
  - KEY_WRAP_EXPORT / KEY_UNWRAP_IMPORT - wrap/unwrap key blobs under a wrapping key slot.
  - KEY_SLOT_CLEAR - clear a slot and invalidate its policy.
  - KEY_SLOT_RDWR - read or write raw key material for SW_KEY slots.
  - SET_TPKEY - install a slot as the active transport protection key.

Design notes
--------------
Commands are modelled as "transactions" - struct hwkm_transaction - that carry both the request and the response. Transactions are queued on a handle - struct hwkm_handle - and executed in FIFO order by hwkm_run_cmd_queue(). This makes multi-step sequences (e.g. clear + derive + read + clear for HUK derivation) easy to express.

Footnote
----------
Only the KM_MASTER destination is currently supported. Slave instances (e.g. GPCE or ICE) can be added by extending hwkm_key_destination and run_transaction().


b49020:
@T0nyJH fyi..

core/drivers/crypto/hwkm/hwkm.c
```diff
@@ -0,0 +1,372 @@
...
+ * context. Subsequent calls return the cached value.
+ *
+ * Slow path:
+ *   Phase 1: Clear the scratch slot, then derive the HUK into it.
+ *   Phase 2: Read the derived HUK from the scratch slot, then clear the
+ *            scratch slot again.
+ *
+ * Return: TEE_SUCCESS on success, or a TEE_ERROR_* code on failure.
+ */
+TEE_Result tee_otp_get_hw_unique_key(struct tee_hw_unique_key *hwkey)
```
ldts, line 136: instead of repeating the same pattern for each phase, is it possible that we can extract the pattern into a static function?

b49020:
Please rebase to the tip of qcom-next as it has been rebased to tip of upstream master to include the merged PAS patch-set as well as latest fuse driver patch-set from Selvam.

core/drivers/crypto/hwkm/hwkm.c
qc-azarrabi: Refactored.

core/arch/arm/plat-qcom/hoya/kodiak/target_config.h
```diff
@@ -43,4 +43,7 @@
 #define PAS_ID_VENUS			9
 #define PAS_ID_TURING			18
 
+#define HWKM_MASTER_BASE	        UL(0x010c0000)
+#define HWKM_MASTER_SIZE	        UL(0x00020000)
```
b49020, line 47: Since the base addresses seems to be common for Hoya architecture, can you rather move the defines to `core/arch/arm/plat-qcom/hoya/arch_config.h`?

qc-azarrabi: will do.

core/drivers/crypto/hwkm/hwkm.c
```diff
@@ -0,0 +1,403 @@
...
+		return 0;
+
+	return (vaddr_t)va;
+}
+
+/**
+ * hwkm_init() - Initialize the HWKM instance.
+ *
+ * Initialization sequence:
+ *   1. Enable the four GCC clock branches required by the HWKM Master.
```
harshaldev27, line 325: Are we yet to add support for enabling the clocks? I couldn't find them in this PR.

core/arch/arm/plat-qcom/hoya/lemans/target.mk
```diff
@@ -9,3 +9,5 @@ _qcom_fuseprov_deps = $(if $(filter y,$(CFG_QCOM_QFPROM_FUSEPROV)),y,n)
 CFG_QCOM_CMD_DB ?= $(_qcom_fuseprov_deps)
 CFG_QCOM_RPMH_CLIENT ?= $(_qcom_fuseprov_deps)
 CFG_QCOM_QFPROM ?= $(_qcom_fuseprov_deps)
+
+CFG_QCOM_HWKM ?= y
```
harshaldev27, line 13: Any reason to omit enabling this driver for Kodiak? As per my understanding, it also has the same HWKM.

core/drivers/crypto/hwkm/hwkm.c
```diff
@@ -0,0 +1,403 @@
...
+
+	t_read->cmd.op = HWKM_OP_KEY_SLOT_RDWR;
+	t_read->cmd.rdwr.slot = HWKM_SCRATCH_SLOT_B;
+	t_read->cmd.rdwr.is_write = false;
+
+	rc = hwkm_run_transactions(HWKM_KEY_DEST_KM_MASTER, 3,
+				   (struct hwkm_transaction *const[]){
+					t_read, t_clear_a, t_clear_b });
+	if (rc) {
+		res = hwkm_to_tee(rc);
```
harshaldev27, line 261: nit: Maybe hwkm_err_to_tee_err()? This confused me on first read. 😄

PR #17, opened 2026-06-26
drivers: qcom: Add TME-Lite IPC stack and enable on Bobcat (ipq96xx/ipq52xx)

zelvam95:
## Summary

This series adds the TME-Lite (TMEL) communication stack to OP-TEE and enables it
on the Bobcat family (ipq96xx, ipq52xx).

TME-Lite is the hardware Root of Trust on Bobcat SoCs — a dedicated RISC-V security
processor, separate from the application cores OP-TEE runs on. OP-TEE (running in
the ARM secure world) talks to TME-Lite over an IPC channel to request
security-sensitive services such as hardware entropy and key derivation, rather than
implementing them locally.

The stack is built bottom-up over five driver commits (each a no-op until platform
enablement), followed by platform enablement and the HUK provider:

crypto_rng / tee_otp_*  (OP-TEE core hooks)
│
tmel_rng / tmel_km / qcom_huk   (feature clients + HUK)
│
tmel_com                         (TME-Lite IPC client)
│
glink_lite                       (G-Link Lite protocol)
│
xport_qmp                        (QMP mailbox transport / MMIO + IRQ)
│
── IPC ──▶ TME-Lite (RISC-V Root of Trust)

## Commits

1. **glink_lite: Add protocol driver** — G-Link Lite, the lightweight G-Link protocol
   engine that exchanges messages with TME-Lite. Protocol layer, no direct HW access.
2. **xport_qmp: Add mailbox transport** — the HW-facing QMP transport behind G-Link
   Lite: programs the QMP mailbox MMIO windows, services the inbound TME→APSS
   interrupt, rings the outbound APSS→TME doorbell.
3. **tmel_com: Add communication client** — the TME-Lite IPC client that marshals
   request/response messages for higher-level clients; uses a static cache-coherent
   buffer carved out of the end of TZDRAM.
4. **tmel_rng: Add TME-Lite RNG client** — requests HW entropy from TME-Lite and
   provides `hw_get_random_bytes()`. Falls back to a U-Boot-SPL-populated IMEM entropy
   pool when TME-Lite is bypassed or native interrupts are masked, so boot-critical RNG
   always succeeds.
5. **tmel_km: Add key management client** — KDF-based key derivation, key distribution
   to crypto engines (ICE/TCSR), and key-handle release over IPC.
6. **plat-qcom: bobcat: Enable TME-Lite stack** — turns the stack on for the Bobcat
   family: per-SoC QMP mailbox/IRQ config, the TZDRAM IPC carveout, and the TCSR fuse
   register macros. HWRNG via TME-Lite is enabled on ipq96xx only.
7. **drivers: qcom: Add TME-Lite HUK provider** — overrides the weak
   `tee_otp_get_hw_unique_key()` / `tee_otp_get_die_id()` hooks, deriving the HUK
   through TME-Lite KM (NIST KDF → distribute to TCSR → read back). The KM client's
   first in-tree consumer. Follows the standalone `<vendor>_huk.c` convention.

## Configuration

- `CFG_QCOM_TMEL_COM` — communication stack (glink_lite + xport_qmp + tmel_com)
- `CFG_QCOM_TMEL_RNG` — RNG client (forced on by `CFG_HWRNG_PTA`, ipq96xx)
- `CFG_QCOM_TMEL_KM` — key management client
- `CFG_QCOM_HUK` — HUK provider (forces `CFG_QCOM_TMEL_KM` on)

All default on for Bobcat; the hoya family (kodiak/lemans) is untouched and the stack
stays off there.

## Testing

Built clean with `arm-gnu-toolchain-14.2.rel1`:

| Platform | Result | Notes |
|----------|--------|-------|
| `qcom-ipq96xx` | ✅ | full TMEL stack + KM + HUK + HWRNG |
| `qcom-ipq52xx` | ✅ | TMEL stack + KM + HUK (HWRNG off) |
| `qcom-lemans`  | ✅ | hoya, regression — unchanged |
| `qcom-kodiak`  | ✅ | hoya, regression — unchanged |

Verified the TMEL/HUK objects link into the ipq96xx image and that
`tee_otp_get_hw_unique_key` / `tee_otp_get_die_id` override the weak core hooks.


core/drivers/qcom/glink_lite/glink_api_lite.c
```diff
@@ -0,0 +1,511 @@
...
+glinki_find_xport_by_name(const char *remote_ss)
+{
+	struct glink_transport_if *if_ptr = NULL;
+	uint32_t exceptions = 0;
+	uint32_t idx = 0;
+
+	if (!remote_ss)
+		return NULL;
+
+	exceptions = cpu_spin_lock_xsave(&glink_cmn_spinlock);
```
b49020, line 36: Why aren't you using simple mutex here instead of CPU spin locks which are expensive?

core/drivers/qcom/glink_lite/glink_com.h
```diff
@@ -0,0 +1,123 @@
...
+ */
+struct glink_channel_ctx;
+struct glink_link_notify;
+
+struct glink_link_info {
+	const char *remote_ss;
+	enum glink_link_state_type link_state;
+};
+
+typedef void (*glink_rx_notification_cb)(struct glink_channel_ctx *handle,
```
b49020, line 47: Please avoid any new typedefs

core/drivers/qcom/sub.mk
```diff
@@ -12,3 +12,6 @@ $(eval $(call cfg-depends-all,CFG_QCOM_RPMH_CLIENT,CFG_QCOM_CMD_DB))
 subdirs-$(CFG_QCOM_CMD_DB) += cmd_db
 subdirs-$(CFG_QCOM_RPMH_CLIENT) += rpmh
 subdirs-$(CFG_QCOM_QFPROM) += qfprom
+
+# TME-Lite Communication stack
+subdirs-$(CFG_QCOM_TMEL_COM) += glink_lite
```
b49020: Rather depend on a `CFG_QCOM_GLINK_LITE`, there might be future users.

core/drivers/qcom/glink_lite/glink_api_lite.c
b49020: Please check this throughout this and other drivers.

core/drivers/qcom/glink_lite/glink_api_lite.c
b49020, line 1: Is this driver used for TME communications on wildcat targets too?

core/drivers/qcom/xport_qmp/sub.mk
b49020, line 1: You can drop these redundant lines from commit message:

```
Port the QMP (Qualcomm Messaging Protocol) transport driver from the
downstream win.optee.1.0 (4.6.0-based) tree.
```

Rather explain the purpose of this driver.

core/arch/arm/plat-qcom/bobcat/ipq52xx/target.mk
```diff
@@ -5,4 +5,9 @@ CFG_QCOM_DIAG_LOG ?= $(if $(filter y,$(CFG_TEE_CORE_DEBUG)),y,n)
 
 CFG_TZDRAM_START ?= 0x87D80000
 CFG_TZDRAM_SIZE ?= 0x280000
+# Reserve the TMEL IPC coherent carveout at the end of TZDRAM.
+ifeq ($(CFG_QCOM_TMEL_COM),y)
+CFG_TEE_RAM_VA_SIZE ?= (0x280000 - $(CFG_TMECOM_IPCBUF_CARVEOUT_SIZE))
```
ldts, line 10: do we need to check for out of bounds

core/drivers/qcom/qcom_huk/sub.mk
```diff
@@ -0,0 +1,6 @@
+# SPDX-License-Identifier: BSD-2-Clause
+#
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+#
+
+srcs-$(CFG_QCOM_HUK) += qcom_huk.c
```
ldts: this will confuse wildcat with the HWKM huk. Try to specialize the name of this QCOM HUK so it is not generic.

core/drivers/qcom/glink_lite/glink_api_lite.c
ldts: also please make sure the mutex is not held unnecesarily - ie for input param validation

core/drivers/qcom/glink_lite/glink_api_lite.c
```diff
@@ -0,0 +1,511 @@
...
+	struct glink_core_xport_ctx *core_priv = &if_ptr->core_priv;
+	struct glink_channel_ctx *ch_ctx = &core_priv->ch_ctx;
+	glink_rx_notification_cb rx_cb = NULL;
+	uint32_t exceptions = 0;
+
+	exceptions = cpu_spin_lock_xsave(&core_priv->cs);
+
+	if (!ch_ctx->if_ptr || !ptr || !size) {
+		cpu_spin_unlock_xrestore(&core_priv->cs, exceptions);
+		return;
```
ldts, line 505: no error handling? check through the PR

ldts (changes requested):

core/drivers/qcom/glink_lite/glink_api_lite.c
zelvam95: @b49020, As per my understanding, mutex can't be used here because part of this locking runs in interrupt context. In OP-TEE, mutex_lock() (core/kernel/mutex.c) is a sleeping lock — on contention it calls wq_wait_final() to sleep the calling thread via the normal world — and it asserts it's running on a thread in normal mode:

```
assert_have_no_spinlock();
assert(thread_get_id_may_fail() != THREAD_ID_INVALID);
assert(thread_is_in_normal_mode());
```
xport_qmp_isr() is a registered interrupt handler; it takes ctx->cs and its handlers call into the G-Link core (glink_core_rx_cmd_*), which take the same locks as the thread-context API. There's no thread to sleep in an ISR, so a mutex is invalid — cpu_spin_lock_xsave() (busy-wait, foreign interrupts masked) is the correct primitive for state shared with an ISR.

This matches existing practice: NXP CAAM (caam_jr.c), ST EXTI/RTC (stm32_exti.c, stm32_rtc.c) and AMD ASU (asu_main.c) all take cpu_spin_lock in their interrupt handlers; no in-tree driver uses a mutex in an ISR. I audited all three drivers here — each shares state with the ISR/masked path, so spinlocks are correct throughout, and I keep the critical sections minimal (see #2).

Let me know if you have any comments/suggestions.

zelvam95: @ldts - Audited the lock scopes across all three drivers, not just one function. Input
validation is done before taking the lock in every public entry point:
`glink_close` / `glink_tx` / `glink_rx_done` / `glink_register_link_state_cb`
return `GLINK_STATUS_INVALID_PARAM` before locking, and
`tmecom_client_get_coherent_buf` / `send_message` validate their args before
`cpu_spin_trylock`. I found one case (`glink_core_rx_cmd_data`) that checked its
`ptr` / `size` args inside the lock and moved those out; **the lock there now
covers only the shared-state check and update.**

Note that some in-lock conditionals read *shared state* rather than input — e.g.
`!ch_ctx->if_ptr` (is the channel open?) and `ctx->state != LINK_UP` (link
state). Those must stay under the lock because they race with the ISR; they are
not parameter validation.

core/drivers/qcom/glink_lite/glink_api_lite.c
zelvam95: The `glink_core_rx_cmd_*` functions are `void` notification entry points invoked
by the transport on the RX path in interrupt context (the ISR returns
`ITRR_HANDLED`) — there is no caller to return an error to. On invalid input they
drop the packet rather than forward a bad buffer. This is now explicit:

```c
if (!ptr || !size)
        return;

exceptions = cpu_spin_lock_xsave(&core_priv->cs);

/* RX path has no caller to return errors to; drop if channel closed */
if (!ch_ctx->if_ptr) {
        cpu_spin_unlock_xrestore(&core_priv->cs, exceptions);
        EMSG("Dropping RX data: channel not open");
        return;
}
```

Functions that can report failure return `enum glink_err_type` / `TEE_Result`
and are checked by their callers.

core/drivers/qcom/glink_lite/glink_com.h
zelvam95: The "no new typedefs" rule (Linux coding style) targets typedefs of structs,
enums and plain scalars. Function-pointer typedefs are the documented exception —
checkpatch's own NEW_TYPEDEFS check explicitly skips them:

```perl
# check for new typedefs, only function parameters and sparse annotations
# make sense.
if ($line =~ /\btypedef\s/ &&
    $line !~ /\btypedef\s+$Type\s*\(\s*\*?$Ident\s*\)\s*\(/ &&  # fn-ptr typedef: skipped
    ...)
        WARN("NEW_TYPEDEFS", ...)
```

and OP-TEE core uses ~25 of them (`itr_handler_t`, `pm_callback`,
`dt_driver_probe_func`, `clk_dt_get_func`, `scmi_msg_handler_t`, ...). I verified
in patch mode that checkpatch flags a `struct`/scalar typedef but not a
function-pointer one.

So I removed the one typedef that the rule does target — `tme_key_handle_t`
(a scalar alias for `uint32_t`), now plain `uint32_t` — and kept the
function-pointer typedefs for the glink callback/transport-op types and
`tmecom_notify_rx_callback`, since inlining them made the struct members and
local snapshots much harder to read. The series is checkpatch-clean (no
NEW_TYPEDEFS warnings).

core/drivers/qcom/sub.mk
zelvam95: Done. `glink_lite` is gated on `CFG_QCOM_GLINK_LITE` and `xport_qmp` on
`CFG_QCOM_XPORT_QMP`, each independently usable; `CFG_QCOM_TMEL_COM`
force-selects both:

```make
subdirs-$(CFG_QCOM_GLINK_LITE) += glink_lite
subdirs-$(CFG_QCOM_XPORT_QMP)  += xport_qmp
```

core/drivers/qcom/glink_lite/glink_api_lite.c
zelvam95: Not directly. Wildcat uses TME-M (this is TME-L). Its stack is similar but built
on the full G-Link driver rather than G-Link Lite, with some protocol
differences, so reusing this would need some rework/abstraction — the wildcat
team can evaluate that on top of this. This series is TME-L only, and the
configs/HUK are named TME-L specific to avoid colliding with the wildcat
mechanisms.

core/drivers/qcom/xport_qmp/sub.mk
zelvam95: Done. Rewrote all 7 commit messages to describe each driver's purpose; verified
no "downstream / win.optee / 4.6.0 / port the" framing remains in any commit
message.

core/arch/arm/plat-qcom/bobcat/ipq52xx/target.mk
zelvam95: The carveout is taken from the top of TZDRAM and TEE RAM is shrunk by the same
amount within the fixed `CFG_TZDRAM_SIZE`, so they cannot overlap by
construction, and the values are compile-time constants (no runtime index). To
guard against a misconfiguration I added a compile-time check in
`tmecom_client.c`:

```c
_Static_assert(TMECOM_IPCBUF_CARVEOUT_SIZE < CFG_TZDRAM_SIZE,
               "TMECOM carveout must fit within TZDRAM");
```

core/drivers/qcom/qcom_huk/sub.mk
zelvam95: Done. Renamed to be TME-L specific: `qcom_huk/` -> `tmel_huk/`,
`qcom_huk.c` -> `tmel_huk.c`, `CFG_QCOM_HUK` -> `CFG_QCOM_TMEL_HUK` (verified no
`qcom_huk` / `CFG_QCOM_HUK` references remain). It is now clearly distinct from
wildcat's HWKM HUK.

PR #19, opened 2026-07-09
pta/ta: qcom: pas: verify PIL firmware segment integrity before reset

zelvam95:
Adds the first layer of PIL firmware image authentication for the PAS
subsystem: the PTA re-hashes each loaded segment against the image's own
MBN hash table before releasing the peripheral from reset, so a
compromised REE cannot substitute firmware after the image metadata has
been accepted. Enabled on Lemans via `CFG_QCOM_PAS_AUTH`.

Signature authentication (certificate chain, device binding,
anti-rollback) is not part of this series; it is added on top of this in a
stacked follow-up, PR #20.

## Design

The implementation splits along the TEE trust boundary:

- **Segment hashing** lives in the PAS pseudo-TA (`pas_auth_core.c`), which
  can call `core_mmu_add_mapping()` to map the non-secure carveout. A user-TA
  cannot do this.

- **Metadata handling** lives in the `qcom_pas` user-TA (`pas_auth.c`), which
  saves a TEE-private copy of the INIT_IMAGE metadata blob and parses the MBN
  hash segment to extract the per-segment digest table.

The flow after this series:

```mermaid
sequenceDiagram
    participant REE as REE (Linux)
    participant TA as qcom_pas TA
    participant PTA as PAS PTA

    Note over REE,PTA: INIT_IMAGE - save metadata + MBN parse
    REE->>TA: INIT_IMAGE(pas_id, metadata_blob)
    TA->>PTA: PTA_QCOM_PAS_INIT_IMAGE
    PTA-->>TA: TEE_SUCCESS
    TA->>TA: pas_auth_save_metadata()<br/>TEE-private copy keyed by pas_id
    TA->>TA: pas_auth_authenticate()<br/>pas_mbn_parse() extracts hash_table
    Note right of TA: MBN parse only, no signature auth in this series
    TA-->>REE: TEE_SUCCESS

    Note over REE,PTA: MEM_SETUP - record carveout coordinates
    REE->>TA: MEM_SETUP(pas_id, fw_base, fw_size)
    TA->>PTA: PTA_QCOM_PAS_MEM_SETUP
    PTA-->>TA: TEE_SUCCESS
    TA-->>REE: TEE_SUCCESS

    Note over REE,PTA: REE loads segments into carveout, no TEE involvement

    Note over REE,PTA: AUTH_AND_RESET - segment hash verification + release
    REE->>TA: AUTH_AND_RESET(pas_id, fw_base, fw_size)
    TA->>PTA: PTA_QCOM_PAS_VERIFY_IMAGE with metadata and hash_table
    PTA->>PTA: map carveout MEM_AREA_RAM_NSEC
    loop for each PT_LOAD segment
        PTA->>PTA: SHA-256/384 then consttime_memcmp vs hash_table entry
    end
    PTA->>PTA: unmap carveout
    PTA-->>TA: TEE_SUCCESS all segments verified
    TA->>PTA: PTA_QCOM_PAS_AUTH_AND_RESET
    PTA-->>TA: TEE_SUCCESS peripheral released from reset
    TA-->>REE: TEE_SUCCESS
```

> If any segment digest mismatches, `PTA_QCOM_PAS_VERIFY_IMAGE` returns
> `TEE_ERROR_SECURITY` and `PTA_QCOM_PAS_AUTH_AND_RESET` is **never invoked**;
> the peripheral stays in reset.

> Signature authentication (cert chain, root-of-trust, SW/HW bindings,
> anti-rollback) is **not present in this series**; it is added in the stacked
> follow-up PR #20.

## Testing

- Built `PLATFORM=qcom-lemans CFG_QCOM_PAS_AUTH=y CFG_WERROR=y`; checkpatch
  clean across the series.
- On a Lemans board: valid, correctly-hashed images load and the peripheral is
  released from reset; an image with a corrupted/mismatched segment digest is
  rejected at `AUTH_AND_RESET` and the peripheral stays in reset, with no side
  effects on subsequent loads.

### DIFF
--- a/core/arch/arm/plat-qcom/hoya/lemans/target.mk
+++ b/core/arch/arm/plat-qcom/hoya/lemans/target.mk
@@ -16,10 +16,11 @@ endif
 CFG_QCOM_PAS_PTA ?= y
 
 ifeq ($(CFG_QCOM_PAS_PTA),y)
-# PAS subsystems map their controller windows at runtime from the reserved VA
-# pool (never released). The six DSP windows total ~146.5 MB; the 60 MB default
-# fits only one, so reserve 256 MB with headroom.
 CFG_RESERVED_VASPACE_SIZE ?= (256 * 1024 * 1024)
 CFG_IN_TREE_EARLY_TAS += qcom_pas/cff7d191-7ca0-4784-af13-48223b9a4fbe
+
+CFG_QCOM_PAS_AUTH ?= y
+
+CFG_PAS_MD_SLOTS ?= 8
 endif
 CFG_QCOM_HWKM ?= y
--- a/core/pta/qcom/pas/pas_auth_core.c
+++ b/core/pta/qcom/pas/pas_auth_core.c
@@ -1,0 +1,403 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2026, Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#include <elf32.h>
+#include <elf64.h>
+#include <kernel/cache_helpers.h>
+#include <mm/core_mmu.h>
+#include <pas_auth_core.h>
+#include <platform_pas.h>
+#include <string.h>
+#include <string_ext.h>
+#include <tee/tee_cryp_utl.h>
+#include <trace.h>
+#include <utee_defines.h>
+#include <util.h>
+
+#include "pas_subsys.h"
+
+#define MI_PBT_PAGE_MODE_MASK		0x00100000
+#define MI_PBT_PAGE_MODE_SHIFT		20
+#define MI_PBT_ACCESS_TYPE_MASK		0x00E00000
+#define MI_PBT_ACCESS_TYPE_SHIFT	21
+#define MI_PBT_SEGMENT_TYPE_MASK	0x07000000
+#define MI_PBT_SEGMENT_TYPE_SHIFT	24
+
+#define MI_PBT_NON_PAGED_SEGMENT	0x0
+#define MI_PBT_HASH_SEGMENT		0x2
+#define MI_PBT_NOTUSED_SEGMENT		0x3
+#define MI_PBT_SHARED_SEGMENT		0x4
+
+#define MI_PBT_PAGE_MODE(x) \
+	(((x) & MI_PBT_PAGE_MODE_MASK) >> MI_PBT_PAGE_MODE_SHIFT)
+#define MI_PBT_ACCESS_TYPE(x) \
+	(((x) & MI_PBT_ACCESS_TYPE_MASK) >> MI_PBT_ACCESS_TYPE_SHIFT)
+#define MI_PBT_SEGMENT_TYPE(x) \
+	(((x) & MI_PBT_SEGMENT_TYPE_MASK) >> MI_PBT_SEGMENT_TYPE_SHIFT)
+
+struct elf_info {
+	size_t elf_hdr_len;
+	size_t phdr_offset;
+	size_t phdr_entry_len;
+	size_t phdr_count;
+	uint64_t entry;
+	bool is_64;
+};
+
+struct phdr_info {
+	uint64_t paddr;
+	size_t file_len;
+	size_t mem_len;
+	uint32_t type;
+	uint32_t flags;
+};
+
+#define PARSE_EHDR(e_info, ehdr, _is_64) do {				\
+	(e_info)->is_64 = (_is_64);					\
+	(e_info)->elf_hdr_len = (ehdr)->e_ehsize;			\
+	(e_info)->phdr_offset = (ehdr)->e_phoff;			\
+	(e_info)->phdr_entry_len = (ehdr)->e_phentsize;		\
+	(e_info)->phdr_count = (ehdr)->e_phnum;			\
+	(e_info)->entry = (ehdr)->e_entry;			\
+} while (0)
+
+#define PARSE_PHDR(p_info, phdr) do {					\
+	(p_info)->paddr = (phdr)->p_paddr;				\
+	(p_info)->file_len = (phdr)->p_filesz;				\
+	(p_info)->mem_len = (phdr)->p_memsz;				\
+	(p_info)->type = (phdr)->p_type;				\
+	(p_info)->flags = (phdr)->p_flags;				\
+} while (0)
+
+static bool check_range(uint64_t off, uint64_t len, uint64_t total)
+{
+	uint64_t end = 0;
+
+	if (ADD_OVERFLOW(off, len, &end))
+		return false;
+
+	return end <= total;
+}
+
+static bool is_hashed(uint32_t p_type, uint32_t p_flags)
+{
+	if (p_type != PT_LOAD)
+		return false;
+
+	return MI_PBT_PAGE_MODE(p_flags) == MI_PBT_NON_PAGED_SEGMENT &&
+	       MI_PBT_SEGMENT_TYPE(p_flags) != MI_PBT_HASH_SEGMENT &&
+	       MI_PBT_ACCESS_TYPE(p_flags) != MI_PBT_NOTUSED_SEGMENT &&
+	       MI_PBT_ACCESS_TYPE(p_flags) != MI_PBT_SHARED_SEGMENT;
+}
+
+static TEE_Result parse_elf(const uint8_t *fw, size_t fw_size,
+			    struct elf_info *e_info)
+{
+	const unsigned char *ident = fw;
+
+	if (fw_size < EI_NIDENT)
+		return TEE_ERROR_BAD_FORMAT;
+
+	if (ident[EI_MAG0] != ELFMAG0 || ident[EI_MAG1] != ELFMAG1 ||
+	    ident[EI_MAG2] != ELFMAG2 || ident[EI_MAG3] != ELFMAG3)
+		return TEE_ERROR_BAD_FORMAT;
+
+	switch (ident[EI_CLASS]) {
+	case ELFCLASS64: {
+		const Elf64_Ehdr *ehdr = (const void *)fw;
+
+		if (fw_size < sizeof(*ehdr))
+			return TEE_ERROR_BAD_FORMAT;
+
+		PARSE_EHDR(e_info, ehdr, true);
+		if (e_info->elf_hdr_len < sizeof(*ehdr) ||
+		    e_info->phdr_entry_len < sizeof(Elf64_Phdr))
+			return TEE_ERROR_BAD_FORMAT;
+		break;
+	}
+	case ELFCLASS32: {
+		const Elf32_Ehdr *ehdr = (const void *)fw;
+
+		if (fw_size < sizeof(*ehdr))
+			return TEE_ERROR_BAD_FORMAT;
+
+		PARSE_EHDR(e_info, ehdr, false);
+		if (e_info->elf_hdr_len < sizeof(*ehdr) ||
+		    e_info->phdr_entry_len < sizeof(Elf32_Phdr))
+			return TEE_ERROR_BAD_FORMAT;
+		break;
+	}
+	default:
+		return TEE_ERROR_BAD_FORMAT;
+	}
+
+	return TEE_SUCCESS;
+}
+
+static void get_phdr(const uint8_t *fw, const struct elf_info *e_info,
+		     size_t idx, struct phdr_info *p_info)
+{
+	const uint8_t *p = fw + e_info->phdr_offset +
+			   idx * e_info->phdr_entry_len;
+
+	if (e_info->is_64) {
+		const Elf64_Phdr *phdr = (const void *)p;
+
+		PARSE_PHDR(p_info, phdr);
+	} else {
+		const Elf32_Phdr *phdr = (const void *)p;
+
+		PARSE_PHDR(p_info, phdr);
+	}
+}
+
+static TEE_Result image_range(const uint8_t *fw, const struct elf_info *e_info,
+			      uint64_t *base, uint64_t *end)
+{
+	uint64_t min_paddr = UINT64_MAX;
+	struct phdr_info p_info = { };
+	uint64_t seg_end = 0;
+	uint64_t max_end = 0;
+	size_t i = 0;
+
+	for (i = 0; i < e_info->phdr_count; i++) {
+		get_phdr(fw, e_info, i, &p_info);
+
+		if (!is_hashed(p_info.type, p_info.flags))
+			continue;
+
+		if (ADD_OVERFLOW(p_info.paddr, p_info.mem_len, &seg_end))
+			return TEE_ERROR_BAD_FORMAT;
+
+		if (p_info.paddr < min_paddr)
+			min_paddr = p_info.paddr;
+		if (seg_end > max_end)
+			max_end = seg_end;
+	}
+
+	if (min_paddr == UINT64_MAX)
+		return TEE_ERROR_BAD_FORMAT;
+
+	*base = ROUNDDOWN(min_paddr, SMALL_PAGE_SIZE);
+	if (ROUNDUP_OVERFLOW(max_end, SMALL_PAGE_SIZE, end))
+		return TEE_ERROR_BAD_FORMAT;
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result hash_verify(uint32_t algo, const uint8_t *data,
+			      size_t data_len, const uint8_t *expected,
+			      size_t hash_len)
+{
+	uint8_t digest[PAS_AUTH_CORE_MAX_HASH_SIZE] = { };
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	if (hash_len > sizeof(digest))
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	res = tee_hash_createdigest(algo, data, data_len, digest, hash_len);
+	if (res)
+		return res;
+
+	if (consttime_memcmp(digest, expected, hash_len))
+		res = TEE_ERROR_SECURITY;
+	else
+		res = TEE_SUCCESS;
+
+	memzero_explicit(digest, sizeof(digest));
+
+	return res;
+}
+
+static TEE_Result verify_elf_header(const struct pas_auth_core_ctx *ctx,
+				    const struct elf_info *e_info)
+{
+	size_t hdr_len = 0;
+
+	if (MUL_OVERFLOW(e_info->phdr_entry_len, e_info->phdr_count, &hdr_len))
+		return TEE_ERROR_BAD_FORMAT;
+
+	if (ADD_OVERFLOW(hdr_len, e_info->elf_hdr_len, &hdr_len))
+		return TEE_ERROR_BAD_FORMAT;
+
+	if (hdr_len > ctx->metadata_len)
+		return TEE_ERROR_BAD_FORMAT;
+
+	return hash_verify(ctx->hash_algo, ctx->metadata, hdr_len,
+			   ctx->hash_table, ctx->hash_len);
+}
+
+TEE_Result pas_auth_core_verify_segments(const struct pas_auth_core_ctx *ctx)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+	const uint8_t *expected = NULL;
+	struct phdr_info p_info = { };
+	struct elf_info e_info = { };
+	size_t phdr_table_len = 0;
+	uint64_t image_end = 0;
+	uint64_t offset = 0;
+	size_t verified = 0;
+	uint64_t base = 0;
+	size_t i = 0;
+
+	if (!ctx || !ctx->hash_table || !ctx->fw || !ctx->metadata)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (!ctx->hash_len || ctx->hash_len > PAS_AUTH_CORE_MAX_HASH_SIZE)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	res = parse_elf(ctx->metadata, ctx->metadata_len, &e_info);
+	if (res)
+		return res;
+
+	if (MUL_OVERFLOW(e_info.phdr_entry_len, e_info.phdr_count,
+			 &phdr_table_len) ||
+	    !check_range(e_info.phdr_offset, phdr_table_len, ctx->metadata_len))
+		return TEE_ERROR_BAD_FORMAT;
+
+	if (ctx->num_entries != e_info.phdr_count)
+		return TEE_ERROR_SECURITY;
+
+	res = verify_elf_header(ctx, &e_info);
+	if (res) {
+		EMSG("PAS auth: ELF header hash mismatch");
+		return res;
+	}
+
+	res = image_range(ctx->metadata, &e_info, &base, &image_end);
+	if (res)
+		return res;
+
+	if (e_info.entry < base || e_info.entry > image_end) {
+		EMSG("PAS auth: entry point outside image");
+		return TEE_ERROR_BAD_FORMAT;
+	}
+
+	if (image_end - base > ctx->fw_size) {
+		EMSG("PAS auth: image larger than carveout");
+		return TEE_ERROR_BAD_FORMAT;
+	}
+
+	for (i = 0; i < e_info.phdr_count; i++) {
+		get_phdr(ctx->metadata, &e_info, i, &p_info);
+
+		if (!is_hashed(p_info.type, p_info.flags) || !p_info.file_len)
+			continue;
+
+		if (p_info.paddr < base)
+			return TEE_ERROR_BAD_FORMAT;
+		offset = p_info.paddr - base;
+
+		if (!check_range(offset, p_info.file_len, ctx->fw_size))
+			return TEE_ERROR_BAD_FORMAT;
+
+		if (p_info.mem_len < p_info.file_len)
+			return TEE_ERROR_BAD_FORMAT;
+		if (!check_range(offset, p_info.mem_len, ctx->fw_size))
+			return TEE_ERROR_BAD_FORMAT;
+		if (p_info.mem_len > p_info.file_len) {
+			void *zi = ctx->fw + offset + p_info.file_len;
+			size_t zi_len = p_info.mem_len - p_info.file_len;
+
+			memset(zi, 0, zi_len);
+
+			/*
+			 * The REE maps this carveout uncached and the DSP
+			 * fetches it outside the CPU's coherency domain, so the
+			 * zeros must reach DDR before reset release.
+			 */
+			dcache_cleaninv_range(zi, zi_len);
+		}
+
+		expected = ctx->hash_table + i * ctx->hash_len;
+
+		res = hash_verify(ctx->hash_algo, ctx->fw + offset,
+				  p_info.file_len, expected, ctx->hash_len);
+		if (res) {
+			EMSG("PAS auth: segment %zu hash mismatch", i);
+			return res;
+		}
+
+		verified++;
+	}
+
+	if (!verified) {
+		EMSG("PAS auth: no loadable segments were hashed");
+		return TEE_ERROR_SECURITY;
+	}
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result pas_platform_verify_image(uint32_t pas_id,
+				     const struct pas_fw_region *fw,
+				     const struct pas_metadata *metadata,
+				     const struct pas_hash_table *hash)
+{
+	struct qcom_pas_subsys *subsys = pas_lookup(pas_id);
+	struct pas_auth_core_ctx ctx = { };
+	TEE_Result res = TEE_ERROR_GENERIC;
+	struct qcom_pas_data *data = NULL;
+	uint32_t fw_size = 0;
+	void *fw_va = NULL;
+
+	if (!subsys)
+		return TEE_ERROR_NOT_SUPPORTED;
+
+	if (!metadata->data || !metadata->size || !hash->table ||
+	    !hash->len || !fw->size)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (!hash->entry_size || hash->len % hash->entry_size)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	data = &subsys->data;
+
+	if (!data->fw_base || !data->fw_size) {
+		EMSG("PAS auth: no MEM_SETUP for pas_id=%#"PRIx32, pas_id);
+		return TEE_ERROR_BAD_STATE;
+	}
+
+	if (fw->base != data->fw_base) {
+		EMSG("PAS auth: base %#"PRIxPA" != MEM_SETUP %#"PRIxPA,
+		     fw->base, data->fw_base);
+		return TEE_ERROR_SECURITY;
+	}
+	fw_size = data->fw_size;
+
+	fw_va = core_mmu_add_mapping(MEM_AREA_RAM_NSEC, fw->base, fw_size);
+	if (!fw_va) {
+		EMSG("PAS auth: can't map carveout %#"PRIxPA"/%#"PRIx32,
+		     fw->base, fw_size);
+		return TEE_ERROR_GENERIC;
+	}
+
+	switch (hash->entry_size) {
+	case TEE_SHA256_HASH_SIZE:
+		ctx.hash_algo = TEE_ALG_SHA256;
+		break;
+	case TEE_SHA384_HASH_SIZE:
+		ctx.hash_algo = TEE_ALG_SHA384;
+		break;
+	default:
+		res = TEE_ERROR_NOT_SUPPORTED;
+		goto out;
+	}
+
+	ctx.hash_len = hash->entry_size;
+	ctx.hash_table = hash->table;
+	ctx.num_entries = hash->len / hash->entry_size;
+	ctx.metadata = metadata->data;
+	ctx.metadata_len = metadata->size;
+	ctx.fw = fw_va;
+	ctx.fw_size = fw_size;
+
+	res = pas_auth_core_verify_segments(&ctx);
+out:
+	if (core_mmu_remove_mapping(MEM_AREA_RAM_NSEC, fw_va, fw_size))
+		EMSG("PAS auth: failed to unmap carveout");
+
+	return res;
+}
--- a/core/pta/qcom/pas/pas_core.c
+++ b/core/pta/qcom/pas/pas_core.c
@@ -4,7 +4,6 @@
  */
 
 #include <drivers/clk_qcom.h>
-#include <mm/core_memprot.h>
 #include <mm/core_mmu.h>
 #include <platform_pas.h>
 #include <trace.h>
@@ -12,7 +11,7 @@
 
 #include "pas_subsys.h"
 
-static struct qcom_pas_subsys *pas_lookup(uint32_t pas_id)
+struct qcom_pas_subsys *pas_lookup(uint32_t pas_id)
 {
 	struct qcom_pas_subsys *subsys = NULL;
 	size_t count = 0;
@@ -98,8 +97,8 @@ TEE_Result pas_platform_set_remote_state(uint32_t pas_id, uint32_t state)
 TEE_Result pas_platform_auth_and_reset(uint32_t pas_id)
 {
 	struct qcom_pas_subsys *subsys = pas_lookup(pas_id);
-	struct qcom_pas_data *data = NULL;
 	TEE_Result res = TEE_ERROR_GENERIC;
+	struct qcom_pas_data *data = NULL;
 
 	if (!subsys)
 		return TEE_ERROR_NOT_SUPPORTED;
@@ -141,9 +140,16 @@ TEE_Result pas_platform_auth_and_reset(uint32_t pas_id)
 TEE_Result pas_platform_shutdown(uint32_t pas_id)
 {
 	struct qcom_pas_subsys *subsys = pas_lookup(pas_id);
+	TEE_Result res = TEE_ERROR_GENERIC;
 
 	if (!subsys || !subsys->ops->fw_shutdown)
 		return TEE_ERROR_NOT_SUPPORTED;
 
-	return subsys->ops->fw_shutdown(&subsys->data);
+	res = subsys->ops->fw_shutdown(&subsys->data);
+	if (!res) {
+		subsys->data.fw_base = 0;
+		subsys->data.fw_size = 0;
+	}
+
+	return res;
 }
--- a/core/pta/qcom/pas/platform/platform_pas.h
+++ b/core/pta/qcom/pas/platform/platform_pas.h
@@ -7,6 +7,8 @@
 #define PLATFORM_PAS_H
 
 #include <resource_table.h>
+#include <tee_api_types.h>
+#include <types_ext.h>
 
 TEE_Result pas_platform_mem_setup(uint32_t pas_id, uint32_t fw_size,
 				  uint32_t fw_base_low, uint32_t fw_base_high);
@@ -19,5 +21,37 @@ TEE_Result pas_platform_is_supported(uint32_t pas_id);
 TEE_Result pas_platform_capabilities(uint32_t pas_id);
 TEE_Result pas_platform_init_image(uint32_t pas_id);
 TEE_Result pas_platform_shutdown(uint32_t pas_id);
+
+struct pas_fw_region {
+	paddr_t base;
+	uint32_t size;
+};
+
+struct pas_metadata {
+	const uint8_t *data;
+	size_t size;
+};
+
+struct pas_hash_table {
+	const uint8_t *table;
+	size_t len;
+	uint32_t entry_size;
+};
+
+#ifdef CFG_QCOM_PAS_AUTH
+TEE_Result pas_platform_verify_image(uint32_t pas_id,
+				     const struct pas_fw_region *fw,
+				     const struct pas_metadata *metadata,
+				     const struct pas_hash_table *hash);
+#else
+static inline TEE_Result
+pas_platform_verify_image(uint32_t pas_id __unused,
+			  const struct pas_fw_region *fw __unused,
+			  const struct pas_metadata *metadata __unused,
+			  const struct pas_hash_table *hash __unused)
+{
+	return TEE_SUCCESS;
+}
+#endif /* CFG_QCOM_PAS_AUTH */
 
 #endif
--- a/lib/libutee/include/pta_qcom_pas.h
+++ b/lib/libutee/include/pta_qcom_pas.h
@@ -84,4 +84,23 @@
  */
 #define PTA_QCOM_PAS_SHUTDOWN			8
 
-#endif /* __PTA_QCOM_REMOTEPROC_H */
+/*
+ * PAS firmware image integrity verification.
+ *
+ * The caller is responsible for ensuring the hash table is trusted.
+ *
+ * [in]  params[0].value.a:	Unique 32bit remote processor identifier
+ * [in]  params[0].value.b:	Firmware size
+ * [in]  params[1].value.a:	32bit LSB firmware memory address
+ * [in]  params[1].value.b:	32bit MSB firmware memory address
+ * [in]  params[2].memref:	Packed [metadata | hash_table] buffer: the
+ *				INIT_IMAGE metadata blob (ELF header + phdrs)
+ *				followed immediately by the per-segment
+ *				hash table (one digest per phdr)
+ * [in]  params[3].value.a:	Digest size in bytes (32=SHA-256, 48=SHA-384)
+ * [in]  params[3].value.b:	Metadata size, i.e. offset of hash_table
+ *				within params[2].memref
+ */
+#define PTA_QCOM_PAS_VERIFY_IMAGE		9
+
+#endif /* __PTA_QCOM_PAS_H */
--- a/ta/qcom_pas/include/pas_mbn_parser_priv.h
+++ b/ta/qcom_pas/include/pas_mbn_parser_priv.h
@@ -1,0 +1,35 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#ifndef __PAS_MBN_PARSER_PRIV_H
+#define __PAS_MBN_PARSER_PRIV_H
+
+#include <stddef.h>
+#include <stdint.h>
+#include <tee_api_types.h>
+
+/* MBN header field offsets (bytes from hash-segment start) */
+#define MBN_OFF_VERSION		0x04
+#define MBN_OFF_QC_SIG_SIZE	0x08
+#define MBN_OFF_QC_CERT_SIZE	0x0c
+#define MBN_OFF_CODE_SIZE	0x14
+#define MBN_OFF_OEM_SIG_SIZE	0x1c
+#define MBN_OFF_OEM_CERT_SIZE	0x24
+#define MBN_OFF_QC_META_SIZE	0x28
+#define MBN_OFF_OEM_META_SIZE	0x2c
+
+#define MBN_HDR_SIZE_V6		0x30
+
+/* Read a little-endian uint32_t from @p. */
+uint32_t pas_mbn_read_u32(const uint8_t *p);
+
+TEE_Result pas_mbn_locate(const uint8_t *md, size_t md_size,
+			  const uint8_t **seg, size_t *seg_size);
+
+TEE_Result pas_mbn_reserve_region(const uint8_t *segment, size_t segment_size,
+				  size_t *offset, size_t len,
+				  const uint8_t **region, size_t *region_len);
+
+#endif /* __PAS_MBN_PARSER_PRIV_H */
--- a/ta/qcom_pas/include/qcom_pas_priv.h
+++ b/ta/qcom_pas/include/qcom_pas_priv.h
@@ -1,0 +1,29 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#ifndef __QCOM_PAS_PRIV_H
+#define __QCOM_PAS_PRIV_H
+
+#include <pas_mbn_parser.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#define PAS_MD_SLOTS		CFG_PAS_MD_SLOTS
+
+struct pas_md_slot {
+	void *meta_data;
+	size_t meta_data_size;
+	uint32_t pas_id;
+	bool in_use;
+	struct pas_mbn mbn;
+	bool ready;
+};
+
+struct qcom_pas_session {
+	struct pas_md_slot slots[PAS_MD_SLOTS];
+};
+
+#endif /* __QCOM_PAS_PRIV_H */
--- a/ta/qcom_pas/src/pas_auth.c
+++ b/ta/qcom_pas/src/pas_auth.c
@@ -1,0 +1,178 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#include <pas_auth.h>
+#include <pas_mbn_parser.h>
+#include <pas_meta.h>
+#include <pta_qcom_pas.h>
+#include <qcom_pas_priv.h>
+#include <string.h>
+#include <tee_internal_api.h>
+#include <utee_defines.h>
+
+static struct pas_md_slot *get_meta_data_slot(struct qcom_pas_session *s,
+					      uint32_t pas_id)
+{
+	size_t i = 0;
+
+	for (i = 0; i < PAS_MD_SLOTS; i++) {
+		if (s->slots[i].in_use && s->slots[i].pas_id == pas_id)
+			return &s->slots[i];
+	}
+
+	return NULL;
+}
+
+TEE_Result pas_auth_save_metadata(struct qcom_pas_session *s, uint32_t pt,
+				  TEE_Param params[TEE_NUM_PARAMS])
+{
+	const uint32_t exp_pt = TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_INPUT,
+						TEE_PARAM_TYPE_MEMREF_INPUT,
+						TEE_PARAM_TYPE_NONE,
+						TEE_PARAM_TYPE_NONE);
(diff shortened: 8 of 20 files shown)


core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,419 @@
...
+				     const uint8_t *hash_table,
+				     size_t table_len, uint32_t hash_size)
+{
+	struct qcom_pas_subsys *subsys = pas_lookup(pas_id);
+	struct pas_auth_core_ctx ctx = { };
+	TEE_Result res = TEE_ERROR_GENERIC;
+	struct qcom_pas_data *data = NULL;
+	void *fw_va = NULL;
+
+	if (!subsys)
```
b49020, line 346: Shouldn't we check here for `IS_ENABLED(CFG_QCOM_PAS_AUTH)`?

core/pta/qcom/pas/pas_core.c
```diff
@@ -141,9 +138,22 @@ TEE_Result pas_platform_auth_and_reset(uint32_t pas_id)
...
+		 * Drop the cached carveout coordinates so a subsequent load of
+		 * the same subsystem must call MEM_SETUP first: VERIFY_IMAGE
+		 * cross-checks its fw_base argument against these fields and
+		 * would otherwise accept stale values.
+		 */
+		subsys->data.fw_base = 0;
+		subsys->data.fw_size = 0;
+	}
+
+	return res;
```
b49020, line 154: Split these changes as a separate patch as these doesn't directly relate to this patch.

core/pta/qcom/pas/pas_auth_core.c
zelvam95: Not needed — when CFG_QCOM_PAS_AUTH=n, the TA-side stubs in pas_auth.h never call VERIFY_IMAGE at all (pas_auth_verify_reset() compiles to a no-op), so this PTA function is never invoked in a mismatched-free build — it's just dead code, not a runtime path we fall through. The TEE_ERROR_NOT_SUPPORTED stub in platform_pas.h only guards against a build-mismatch corner case (TA built with auth on, core built with it off). Disabling the config cleanly reverts to the plain PTA-side load/reset flow, no failure introduced.

core/pta/qcom/pas/pas_core.c
zelvam95: Done, added it as a separate commit;

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,419 @@
...
+		if (paddr < base)
+			return TEE_ERROR_BAD_FORMAT;
+		offset = paddr - base;
+
+		if (!range_ok(offset, filesz, ctx->fw_size))
+			return TEE_ERROR_BAD_FORMAT;
+
+		expected = ctx->hash_table + i * ctx->hash_size;
+
+		res = hash_verify(ctx->hash_algo, ctx->fw + offset, filesz,
```
TonyJH1: Before calling hash_verify(), some extra checks may be needed.

If memsz > filesz then this is due to the ZI section not being included in the image loaded from the file system, therefore the difference needs to be memset() to ensure that there is no stale data left in the ZI region.

Also worth bounds checking memsz to ensure that it doesn't overflow any carved out region.

ta/qcom_pas/src/pas_mbn_parser.c
```diff
@@ -0,0 +1,216 @@
...
+	memcpy(&v, p, sizeof(v));
+
+	return v;
+}
+
+TEE_Result pas_mbn_locate(const uint8_t *md, size_t md_size,
+			  const uint8_t **seg, size_t *seg_size,
+			  size_t *preamble)
+{
+	const Elf32_Ehdr *ehdr = (const void *)md;
```
TonyJH1: pas_mbn_locate() currently only handles Elf32 class images, whereas parse_elf() accepts both 32 and 64 class headers. So it would seem that the 64 bit paths would not be executed as they would be rejected with BAD_FORMAT by this function.

ta/qcom_pas/src/qcom_pas.c
```diff
@@ -1,29 +1,58 @@
...
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	res = pas_auth_save_metadata(s, pt, params);
+	if (res)
+		return res;
+
+	res = TEE_InvokeTACommand(pta_session, TEE_TIMEOUT_INFINITE,
+				  PTA_QCOM_PAS_INIT_IMAGE, pt, params, NULL);
+	if (res)
+		return res;
```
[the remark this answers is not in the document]
zelvam95: **Addressed.** `qcom_pas_init_image()` now invokes `PTA_QCOM_PAS_INIT_IMAGE`
first and returns on failure **before** `pas_auth_save_metadata()` copies
anything into a slot. A comment above the call spells out the ordering intent
so it isn't accidentally flipped back:

```c
/*
 * Invoke the PTA first, then stash the metadata into a session slot
 * only on success. Doing this the other way round leaks a slot on
 * every failed PTA invocation - repeated failures across distinct
 * pas_id values would eventually exhaust PAS_MD_SLOTS and leak the
 * associated TA heap until the session is torn down.
 */
```
Ref: `ta/qcom_pas/src/qcom_pas.c` in commit
`ta: qcom_pas: verify loaded firmware image integrity`.

---

ta/qcom_pas/src/pas_mbn_parser.c
```diff
@@ -0,0 +1,216 @@
...
+	if (MUL_OVERFLOW(ehdr->e_phentsize, ehdr->e_phnum, &phtab_end) ||
+	    ADD_OVERFLOW(phtab_end, ehdr->e_phoff, &phtab_end) ||
+	    phtab_end > md_size)
+		return TEE_ERROR_BAD_FORMAT;
+
+	phdrs = (const void *)(md + ehdr->e_phoff);
+	preamble_len = phdrs[0].p_filesz;
+
+	if (preamble_len >= md_size)
+		return TEE_ERROR_BAD_FORMAT;
```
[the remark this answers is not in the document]
zelvam95: **Addressed.** `pas_mbn_locate()` was rewritten to no longer use
`phdrs[0].p_filesz` for the preamble length. It now:

1. Bounds-checks the phdr table itself: `phoff + phentsize*phnum <= md_size`
   with overflow-safe arithmetic (`ADD_OVERFLOW`/`MUL_OVERFLOW`).
2. Walks all program headers and picks the one whose `p_flags`
   `MBN_PT_FLAG_TYPE_MASK` bits mark it as the hash segment.
3. Bounds-checks the hash segment's `[p_offset, p_offset + p_filesz)`
   against `md_size` before dereferencing.

The old "trust phdrs[0]" path is gone.

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,419 @@
...
+		info->ehsize = ehdr->e_ehsize;
+		info->phoff = ehdr->e_phoff;
+		info->phentsize = ehdr->e_phentsize;
+		info->phnum = ehdr->e_phnum;
+	}
+
+	if (info->phentsize < phent_size)
+		return TEE_ERROR_BAD_FORMAT;
+
+	return TEE_SUCCESS;
```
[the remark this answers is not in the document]
zelvam95: *Addressed.** `parse_elf()` now rejects up front:

```c
/*
 * A signer that shrinks e_ehsize below the class's real Ehdr size
 * would make verify_elf_header() hash fewer bytes than an unmodified
 * image, silently narrowing entry-0 coverage. Reject up front.
 */
if (info->ehsize < ehdr_size)
    return TEE_ERROR_BAD_FORMAT;
```

`ehdr_size` is `sizeof(Elf64_Ehdr)` or `sizeof(Elf32_Ehdr)` depending on
`EI_CLASS`, so entry-0 coverage cannot be silently narrowed.

lib/libutee/include/pta_qcom_pas.h
```diff
@@ -84,4 +84,25 @@
  */
 #define PTA_QCOM_PAS_SHUTDOWN			8
 
+/*
+ * PAS firmware image integrity verification.
+ *
+ * Recomputes the per-segment digests of the firmware already loaded into its
+ * carveout and compares them to an authenticated hash table. Invoked by the
+ * PAS user TA, which has signature-verified the hash table beforehand.
+ *
```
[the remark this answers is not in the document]
zelvam95: **Addressed — and this drove a wider pass over the PR.** You were right that
the whole PR was written in the finished two-PR voice and overstated what it
delivers. I reworked PR#19 so it claims **only segment-hash verification**
everywhere, and marked the gap with a generic, position-free `TODO`.

The `VERIFY_IMAGE` doc block now describes the caller contract without
implying an authentication step that isn't here:

```
 * Recomputes the per-segment digests of the firmware already loaded into its
 * carveout and compares them to a caller-supplied hash table. The caller is
 * responsible for ensuring the hash table is trusted (e.g. the PAS user TA
 * performs metadata parsing and, when signature authentication is enabled,
 * anchors the table to the fused root-of-trust before invoking this command).
```

The same honesty pass fixed the file-header of `pas_auth.h`, the
`platform_pas.h` / `pas_auth_core.h` `@hash_table` docs (now "per-segment
digest table", not "authenticated"), `user_ta.mk`, `target.mk`, and renamed
the readiness flag (see thread 8/9 replies and the "Field rename" note below).
Signature auth is restored in the stacked PR#20 at the commit that implements
it.

---

ta/qcom_pas/src/pas_mbn_parser.c
zelvam95: **Addressed.** `pas_mbn_locate()` now switches on `EI_CLASS` and handles both
`ELFCLASS32` and `ELFCLASS64` explicitly, using `Elf32_Ehdr`/`Elf32_Phdr`
or `Elf64_Ehdr`/`Elf64_Phdr` respectively. The ELF-class detection is now
consistent between `pas_mbn_locate()` (TA-side) and `parse_elf()`
(core-side); a 64-bit image is no longer rejected by the parser.

core/pta/qcom/pas/pas_auth_core.c
zelvam95: **Addressed.** Both checks are now in `pas_auth_core_verify_segments()`
around line 300 with an explicit comment covering the reasoning:

```c
/*
 * A segment's [filesz, memsz) ZI tail must be inside the
 * carveout and zeroed before hashing so stale REE bytes cannot
 * slip past the digest or run as uninitialized data.
 * memsz < filesz is malformed.
 */
if (memsz < filesz)
    return TEE_ERROR_BAD_FORMAT;
if (!range_ok(offset, memsz, ctx->fw_size))
    return TEE_ERROR_BAD_FORMAT;
if (memsz > filesz)
    memset(ctx->fw + offset + filesz, 0, memsz - filesz);
```

`memsz < filesz` → reject as malformed; `memsz` bounded against `fw_size`
via the same `range_ok()` used for `filesz`; ZI tail `[filesz, memsz)`
zeroed before hashing.

core/arch/arm/plat-qcom/hoya/lemans/target.mk
```diff
@@ -21,5 +21,11 @@ ifeq ($(CFG_QCOM_PAS_PTA),y)
 # fits only one, so reserve 256 MB with headroom.
 CFG_RESERVED_VASPACE_SIZE ?= (256 * 1024 * 1024)
 CFG_IN_TREE_EARLY_TAS += qcom_pas/cff7d191-7ca0-4784-af13-48223b9a4fbe
+
+# Authenticate each PIL firmware image on INIT_IMAGE: verify the per-segment
+# hash table against the loaded firmware, and (on devices with secure-boot
+# fuses blown) validate the image's certificate chain, signature and
+# fuse-bound bindings before releasing the peripheral from reset.
+CFG_QCOM_PAS_AUTH ?= y
```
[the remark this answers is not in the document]
zelvam95: > The PR description calls this feature CFG_QCOM_PAS_HASH_VERIFY (default off),
> but the code enables CFG_QCOM_PAS_AUTH by default on lemans and the surrounding
> comment implies certificate/signature validation that isn't present in this
> series.

**Addressed.** Two problems here, both fixed:

1. **The `target.mk` comment overstated the feature** *and* was attached to
   the wrong phase (it implied cert/signature validation at INIT_IMAGE). It now
   describes only what this PR does — re-hash at AUTH_AND_RESET before reset
   release — with a generic `TODO` for the signature layer:

   ```make
   # Verify PIL firmware at AUTH_AND_RESET: re-hash the segments the REE loaded
   # and compare them against the per-segment hash table in the image metadata,
   # before releasing the peripheral from reset.
   #
   # TODO: signature authentication of that hash table (certificate chain and
   # fuse bindings) will be added incrementally.
   CFG_QCOM_PAS_AUTH ?= y
   ```

2. **`CFG_QCOM_PAS_HASH_VERIFY` never existed in the tree** — that was a
   mistake in the original PR description. There is only `CFG_QCOM_PAS_AUTH`,
   the umbrella knob for both this PR's segment-hash verification and PR#20's
   signature layer. The PR description has been rewritten to match reality:
   PR#19 enables hash verification under `CFG_QCOM_PAS_AUTH`; PR#20 (stacked)
   adds signature + fuse-binding + anti-rollback under the *same* knob;
   `CFG_QCOM_PAS_AUTH ?= y` on Lemans turns the chain on by default and each PR
   describes only the behavior it actually ships.

---

ta/qcom_pas/user_ta.mk
```diff
@@ -1,4 +1,19 @@
...
+else
 CFG_PAS_TA_HEAP_SIZE ?= (4 * 1024)
+endif
+
+# Authenticate each PIL firmware image on INIT_IMAGE: verify the per-segment
+# hash table against the loaded firmware, and (on devices with secure-boot
+# fuses blown) validate the image's certificate chain, signature and
+# fuse-bound bindings before releasing the peripheral from reset. Disable
+# on configurations where the TA is not responsible for image
+# authentication.
```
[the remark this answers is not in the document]
zelvam95: **Addressed.** Both comments in `user_ta.mk` now describe segment-hash
verification only — no certificate/signature claims:

```make
# Verify each PIL firmware image before releasing the peripheral from
# reset: re-hash the loaded segments and compare them against the
# per-segment hash table carried in the image metadata.
CFG_QCOM_PAS_AUTH ?= n
```

The deferred signature work is still marked with a generic, position-free
`TODO`, but on a maintainer request to keep TODO markers from cluttering the
tree it is now carried once in the config/header comments (`target.mk`,
`platform_pas.h`, `pas_auth.h`) rather than repeated here in `user_ta.mk`.

I also **corrected the heap-sizing comment**, which was independently wrong in
this PR: it justified the size with "plus mbedTLS working room", but this PR's
TA never links mbedTLS (only the MBN parser + integrity backend). The 512 KB
heap is now justified by the ~400 KB TEE-private metadata copy alone; the
mbedTLS/X.509 working-memory justification and the larger heap move to PR#20,
where the crypto actually lands.

---

ta/qcom_pas/src/pas_auth.c
```diff
@@ -0,0 +1,204 @@
...
+		return TEE_ERROR_OUT_OF_MEMORY;
+	}
+
+	size = params[1].memref.size;
+	if (size) {
+		copy = TEE_Malloc(size, TEE_MALLOC_FILL_ZERO);
+		if (!copy)
+			return TEE_ERROR_OUT_OF_MEMORY;
+		memcpy(copy, params[1].memref.buffer, size);
+	}
```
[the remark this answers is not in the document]
zelvam95: **Addressed.** Explicit NULL check added before the `memcpy`:

```c
size = params[1].memref.size;
if (size && !params[1].memref.buffer)
    return TEE_ERROR_BAD_PARAMETERS;
if (size) {
    copy = TEE_Malloc(size, TEE_MALLOC_FILL_ZERO);
    ...
    memcpy(copy, params[1].memref.buffer, size);
}
```

Malformed parameters now fail with `TEE_ERROR_BAD_PARAMETERS` rather than
dereferencing NULL.

---

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,399 @@
...
+#include <string.h>
+#include <string_ext.h>
+#include <tee/tee_cryp_utl.h>
+#include <trace.h>
+#include <utee_defines.h>
+#include <util.h>
+
+#include "pas_subsys.h"
+
+/* Bounds check in u64 to avoid narrowing 64-bit ELF fields. */
```
ldts: this comment about how a user choses to use the function adds 0 value

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,399 @@
...
+{
+	uint64_t end = 0;
+
+	if (ADD_OVERFLOW(off, len, &end))
+		return false;
+
+	return end <= total;
+}
+
+#define MI_PBT_PAGE_MODE_MASK		0x00100000
```
ldts, line 21: Defines before functions

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,399 @@
...
+
+static void get_phdr(const uint8_t *fw, const struct elf_info *info,
+		     size_t idx, uint32_t *p_type, uint32_t *p_flags,
+		     uint64_t *p_paddr, size_t *p_filesz, size_t *p_memsz)
+{
+	const uint8_t *p = fw + info->phoff + idx * info->phentsize;
+
+	if (info->is_64) {
+		const Elf64_Phdr *phdr = (const void *)p;
+
```
ldts, line 147: why are we duplicating the code that derrefences the pointer

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,399 @@
...
+		if (paddr < min_paddr)
+			min_paddr = paddr;
+	}
+
+	return min_paddr;
+}
+
+static TEE_Result hash_verify(uint32_t algo, const uint8_t *data,
+			      size_t data_len, const uint8_t *expected,
```
ldts, line 191: data_len and hash_size? hash_len?

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,399 @@
...
+	for (i = 0; i < info.phnum; i++) {
+		get_phdr(ctx->metadata, &info, i, &type, &flags, &paddr,
+			 &filesz, &memsz);
+
+		if (!is_hashed(type, flags) || !filesz)
+			continue;
+
+		if (paddr < base)
+			return TEE_ERROR_BAD_FORMAT;
+		offset = paddr - base;
```
ldts: add a return line

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,399 @@
...
+
+		if (!range_ok(offset, filesz, ctx->fw_size))
```
ldts: range_ok --> check_range

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,399 @@
...
+		 * A segment's [filesz, memsz) ZI tail must be inside the
+		 * carveout and zeroed before hashing so stale REE bytes cannot
+		 * slip past the digest or run as uninitialized data.
+		 * memsz < filesz is malformed.
+		 */
+		if (memsz < filesz)
+			return TEE_ERROR_BAD_FORMAT;
+		if (!range_ok(offset, memsz, ctx->fw_size))
+			return TEE_ERROR_BAD_FORMAT;
+		if (memsz > filesz)
```
ldts: everywhere, memsz/filesz to mem_len/file_len or similar...

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,399 @@
...
+		EMSG("PAS auth: no loadable segments were hashed");
+		return TEE_ERROR_SECURITY;
+	}
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result pas_platform_verify_image(uint32_t pas_id, uint32_t fw_size,
+				     paddr_t fw_base, const uint8_t *metadata,
+				     size_t metadata_size,
```
ldts: metadata_size and table_len is inconsistent...check everwhere

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,399 @@
...
+		return TEE_ERROR_NOT_SUPPORTED;
+
+	if (!metadata || !metadata_size || !hash_table ||
+	    !table_len || !fw_size)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (!hash_size || table_len % hash_size)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	data = &subsys->data;
```
ldts, line 356: do we need to check for null?

core/pta/qcom/pas/pas_core.c
```diff
@@ -141,9 +137,22 @@ TEE_Result pas_platform_auth_and_reset(uint32_t pas_id)
...
 {
 	struct qcom_pas_subsys *subsys = pas_lookup(pas_id);
+	TEE_Result res = TEE_ERROR_GENERIC;
 
 	if (!subsys || !subsys->ops->fw_shutdown)
 		return TEE_ERROR_NOT_SUPPORTED;
 
-	return subsys->ops->fw_shutdown(&subsys->data);
+	res = subsys->ops->fw_shutdown(&subsys->data);
+	if (!res) {
```
ldts, line 149: should we take this action in the fw_shutdown handler instead? seems cleaner to encapsulate on the handler

core/pta/qcom/pas/platform/platform_pas.h
```diff
@@ -20,4 +22,44 @@ TEE_Result pas_platform_capabilities(uint32_t pas_id);
 TEE_Result pas_platform_init_image(uint32_t pas_id);
 TEE_Result pas_platform_shutdown(uint32_t pas_id);
```
ldts, line 24: do we need all these comments?

core/pta/qcom/pas/platform/platform_pas.h
```diff
@@ -20,4 +22,44 @@ TEE_Result pas_platform_capabilities(uint32_t pas_id);
...
+ * @fw_base:		firmware carveout physical base address
+ * @metadata:		ELF metadata blob (header + phdrs); used for ELF
+ *			parsing and ELF header hash (entry 0); NOT in carveout
+ * @metadata_size:	size of @metadata in bytes
+ * @hash_table:		per-segment digest table, one entry per program header
+ * @table_len:		size of @hash_table in bytes
+ * @hash_size:		digest size in bytes (32 for SHA-256, 48 for SHA-384)
+ */
+#ifdef CFG_QCOM_PAS_AUTH
+TEE_Result pas_platform_verify_image(uint32_t pas_id, uint32_t fw_size,
```
ldts: can you reduce the number of parameters to maybe just three structs? fw, metadata and hash?

core/pta/qcom/pas/platform/platform_pas.h
```diff
@@ -20,4 +22,44 @@ TEE_Result pas_platform_capabilities(uint32_t pas_id);
...
+pas_platform_verify_image(uint32_t pas_id __unused,
+			  uint32_t fw_size __unused,
+			  paddr_t fw_base __unused,
+			  const uint8_t *metadata __unused,
+			  size_t metadata_size __unused,
+			  const uint8_t *hash_table __unused,
+			  size_t table_len __unused,
+			  uint32_t hash_size __unused)
+{
+	return TEE_ERROR_NOT_SUPPORTED;
```
ldts: should this error out? or just pass?

ta/qcom_pas/include/pas_mbn_parser_priv.h
```diff
@@ -0,0 +1,61 @@
...
+ * @seg:      hash segment from pas_mbn_locate()
+ * @seg_size: size of @seg in bytes
+ * @cursor:   in/out: byte offset within @seg; advanced by @len on success
+ * @len:      length of the region to slice; 0 yields a NULL/zero-length
+ *            region without moving @cursor
+ * @ptr:      out: pointer to the region, or NULL when @len is 0
+ * @ptr_len:  out: length of the region
+ */
+TEE_Result pas_mbn_take_region(const uint8_t *seg, size_t seg_size,
+			       size_t *cursor, size_t len,
```
ldts: cursor --> offset, ptr --> region, seg-->segment

ta/qcom_pas/include/qcom_pas_priv.h
```diff
@@ -0,0 +1,44 @@
...
+
+#ifndef __QCOM_PAS_PRIV_H
+#define __QCOM_PAS_PRIV_H
+
+#include <pas_mbn_parser.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+/*
```
ldts: maybe remove all these comments and explain the functionality on the OP-TEE docs? it is just that there is way too many of them...if we want a system view we can always ask the AI to explain

ta/qcom_pas/include/qcom_pas_priv.h
```diff
@@ -0,0 +1,44 @@
...
+ * per-session (rather than as a TA global) ensures two concurrent sessions
+ * cannot observe each other's metadata.
+ *
+ * The remoteproc driver opens one TEE session shared by every peripheral,
+ * and DSPs load concurrently. Metadata must therefore be keyed by pas_id to
+ * prevent one DSP's INIT_IMAGE overwriting another's slot before its
+ * AUTH_AND_RESET runs. Hoya platforms expose at most 7 PAS subsystems; size
+ * the table with headroom.
+ */
+#define PAS_MD_SLOTS	8U
```
ldts: should this be platform dependent then? hoya/bobcat/etc?

ta/qcom_pas/include/qcom_pas_priv.h
```diff
@@ -0,0 +1,44 @@
...
+
+struct pas_md_slot {
+	void *md;
```
ldts: md -> meta_data ?

ta/qcom_pas/include/qcom_pas_priv.h
```diff
@@ -0,0 +1,44 @@
...
+	size_t md_size;
+	uint32_t pas_id;
+	bool used;
+	struct pas_mbn mbn;
+	bool hash_table_valid;
```
ldts: hash_table_valid? do we care if it is a hash_table or something else? what is this variable invalidating?

ta/qcom_pas/src/pas_auth.c
```diff
@@ -0,0 +1,156 @@
...
+
+#include <pas_auth.h>
+#include <pas_mbn_parser.h>
+#include <pta_qcom_pas.h>
+#include <qcom_pas_priv.h>
+#include <string.h>
+#include <tee_internal_api.h>
+#include <utee_defines.h>
+
+static struct pas_md_slot *find_md_slot(struct qcom_pas_session *s,
```
ldts: get_meta_data_slot?

ta/qcom_pas/src/pas_auth.c
```diff
@@ -0,0 +1,156 @@
...
+	slot = find_md_slot(s, pas_id);
+	if (!slot) {
+		for (i = 0; i < PAS_MD_SLOTS; i++) {
+			if (!s->md[i].used) {
+				slot = &s->md[i];
+				break;
+			}
+		}
+	}
+	if (!slot) {
```
ldts: not the right place to check, move it above?

ta/qcom_pas/src/pas_auth.c
```diff
@@ -0,0 +1,156 @@
...
+		EMSG("PAS auth: no free md slot (pas_id=%"PRIu32")", pas_id);
+		return TEE_ERROR_OUT_OF_MEMORY;
+	}
+
+	size = params[1].memref.size;
+	if (size) {
+		copy = TEE_Malloc(size, TEE_MALLOC_FILL_ZERO);
```
ldts: copy? can you rename to what it is?

ta/qcom_pas/src/pas_auth.c
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	res = pas_mbn_parse(slot->md, slot->md_size, TEE_SHA384_HASH_SIZE,
+			    &slot->mbn);
+	if (res) {
+		EMSG("PAS auth: MBN parse failed: %#"PRIx32, res);
+		return res;
+	}
+
+	slot->hash_table_valid = true;
```
ldts: slot->ready? or similar?

ta/qcom_pas/src/pas_auth.c
```diff
@@ -0,0 +1,156 @@
...
+TEE_Result pas_auth_verify_reset(struct qcom_pas_session *s,
+				 TEE_TASessionHandle pta_session,
+				 uint32_t pas_id,
+				 TEE_Param params[TEE_NUM_PARAMS])
+{
+	struct pas_md_slot *slot = find_md_slot(s, pas_id);
+	TEE_Param vp[TEE_NUM_PARAMS] = { };
+	TEE_Result res = TEE_ERROR_GENERIC;
+	size_t combined_size = 0;
+	uint8_t *combined = NULL;
```
ldts: combined? maybe just buffer?

ta/qcom_pas/src/pas_mbn_parser.c
```diff
@@ -0,0 +1,287 @@
...
+	memcpy(&v, p, sizeof(v));
+
+	return v;
+}
+
+TEE_Result pas_mbn_locate(const uint8_t *md, size_t md_size,
+			  const uint8_t **seg, size_t *seg_size)
+{
+	const unsigned char *ident = md;
+	size_t phoff = 0;
```
ldts: phoff, phsize etc..

ta/qcom_pas/src/pas_mbn_parser.c
```diff
@@ -0,0 +1,287 @@
...
+		return TEE_ERROR_BAD_FORMAT;
+
+	*seg = md + hash_off;
+	*seg_size = hash_filesz;
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result pas_mbn_take_region(const uint8_t *seg, size_t seg_size,
+			       size_t *cursor, size_t len,
```
ldts: cursor --> offset

ta/qcom_pas/src/pas_mbn_parser.c
```diff
@@ -0,0 +1,287 @@
...
+	    hash_end > md_size || !hash_filesz)
+		return TEE_ERROR_BAD_FORMAT;
+
+	*seg = md + hash_off;
+	*seg_size = hash_filesz;
+
+	return TEE_SUCCESS;
+}
+
+TEE_Result pas_mbn_take_region(const uint8_t *seg, size_t seg_size,
```
ldts: take --> reserve?

ta/qcom_pas/src/pas_mbn_parser.c
```diff
@@ -0,0 +1,287 @@
...
+			       size_t *cursor, size_t len,
+			       const uint8_t **ptr, size_t *ptr_len)
+{
+	if (!len) {
+		*ptr = NULL;
```
ldts: why do we need to update these values?

ldts (changes requested):

ta/qcom_pas/src/sub.mk
```diff
@@ -1 +1,4 @@
+global-incdirs-y += ../include
 srcs-y += qcom_pas.c
+srcs-$(CFG_QCOM_PAS_AUTH) += pas_auth.c
```
ldts: single line?

core/pta/qcom/pas/pas_auth_core.c
zelvam95: Deleted the verbose comment above `range_ok()`. The function name is self-documenting now.

core/pta/qcom/pas/pas_auth_core.c
zelvam95: Moved `range_ok()` (renamed to `check_range()`) below the `MI_PBT_*` macro block.

core/pta/qcom/pas/pas_auth_core.c
zelvam95: Collapsed it — the two branches now share a COPY_PHDR_FIELDS(phdr) helper instead of repeating the five field copies, so only the Elf32_Phdr/Elf64_Phdr cast differs between them.

core/pta/qcom/pas/pas_auth_core.c
zelvam95: hash_size emphasizes it's a property of the hash algorithm (SHA-256 is always 32 bytes)
data_len emphasizes it's a variable length of input data

I hope its fine to leave it as it is? If you want me to rename it to hash_len for uniformity, let me know & I will do that. Above is the rationale for current naming.

core/pta/qcom/pas/pas_auth_core.c
zelvam95: Added blank-line separation before the `return TEE_ERROR_BAD_FORMAT;`

core/pta/qcom/pas/pas_auth_core.c
zelvam95: Renamed all occurrences

core/pta/qcom/pas/pas_auth_core.c
zelvam95: Renamed `metadata_size` → `metadata_len` for consistency.

core/pta/qcom/pas/pas_auth_core.c
zelvam95: updated.

core/pta/qcom/pas/pas_auth_core.c
zelvam95: Yes, and we already do at the top of the func.

```
	if (!subsys) 
		return TEE_ERROR_NOT_SUPPORTED;
```

core/pta/qcom/pas/platform/platform_pas.h
zelvam95: Trimmed the big doc comment from ~15 lines to ~4 lines. Kept: what it does, that the hash table is caller-supplied/not authenticated, and the existing TODO. Dropped the per-parameter `@foo:` list since params are now 3 self-documenting structs.

core/pta/qcom/pas/platform/platform_pas.h
zelvam95: Done     
```
               const struct pas_fw_region *fw,
               const struct pas_metadata *metadata,
               const struct pas_hash_table *hash
```

core/pta/qcom/pas/platform/platform_pas.h
zelvam95: The PTA-side stub returns `TEE_ERROR_NOT_SUPPORTED`, but this is **never reached in practice** because the TA-side stubs (in `ta/qcom_pas/include/pas_auth.h`) silently return `TEE_SUCCESS` when `CFG_QCOM_PAS_AUTH=n`. The full authentication pipeline is stubbed at the TA level, so firmware loads without verification when the config is disabled. The PTA stub is a safety net for direct PTA callers, but the normal flow (REE → TA → PTA) never reaches it.

ta/qcom_pas/include/pas_mbn_parser_priv.h
zelvam95: Updated as below:
Rename cursor → offset
Rename seg/seg_size → segment/segment_size
Rename ptr/ptr_len → region/region_len

ta/qcom_pas/include/qcom_pas_priv.h
zelvam95: Renamed struct field `md` → `meta_data` and `md_size` → `meta_data_size` in `struct pas_md_slot`

ta/qcom_pas/include/qcom_pas_priv.h
zelvam95: If secure boot en -> we do signature verification first and if its successful -> Then hash table is valid -> hash verification is done on that trusted meta data.

In case of non-secure boot -> we don't do signature verification; If we have a valid hash table (basic checks pass) -> Then hash table is set as valid -> hash verification alone is done.

We could have used authenticated or some other flag but that wouldn't make sense in case of non-secure boot. Above was a generic flag which made sense for both secure and non-secure boot and hence kept so.

Let me know if you have a better suggestion.

ta/qcom_pas/src/pas_auth.c
zelvam95: Renamed the function and all call sites.

ta/qcom_pas/src/pas_auth.c
zelvam95: Renamed local `copy` → `meta_data_copy` for clarity.

ta/qcom_pas/src/pas_auth.c
zelvam95: Moved the `size && !buffer` null check from after slot allocation to the top of the function (right after the `pt` check). This validates inputs before doing any work.

ta/qcom_pas/src/pas_auth.c
zelvam95: https://github.com/qualcomm-linux/optee_os/pull/19#discussion_r3741021377

For now have kept as hash_table_valid to indicate meta data is authenticated in case of secure boot/basic checks are done incase of NS boot. Do you think ready makes sense for this purpose? I have shared more dets on exactly what we're trying to store in this variable in the above comment.

ta/qcom_pas/src/pas_auth.c
zelvam95: Renamed `combined`/`combined_size` → `buffer`/`buffer_size`

ta/qcom_pas/src/pas_mbn_parser.c
zelvam95: Renamed `cursor` → `offset`

ta/qcom_pas/src/pas_mbn_parser.c
zelvam95: Renamed `pas_mbn_take_region()` → `pas_mbn_reserve_region()`

ta/qcom_pas/include/qcom_pas_priv.h
zelvam95: Trimmed the big file-header comment from ~15 lines to ~3 lines. Kept: per-session slots keyed by pas_id, TEE-private copy prevents REE tampering.

ta/qcom_pas/src/sub.mk
zelvam95: Collapsed the multi-line `srcs-$(CFG_QCOM_PAS_AUTH) +=` block to a single space-separated assignment.

ta/qcom_pas/src/pas_mbn_parser.c
zelvam95: The hash-segment offset is now computed the way QTEE does — ehdr_size + phentsize * phnum with overflow guards — so no untrusted phdr offset is carried forward.

The flagged lines are gone.

ta/qcom_pas/include/qcom_pas_priv.h
zelvam95: Made it platform-tunable. The slot count now derives from CFG_PAS_MD_SLOTS, defaulting to 1 generically and overridden per target — Lemans sets CFG_PAS_MD_SLOTS ?= 8 in its target.mk to cover the PAS subsystems that load concurrently. So it's per-target now rather than a hardcoded bound.

ta/qcom_pas/src/pas_mbn_parser.c
zelvam95: FYI:
phoff: "where does the table start?" → one address.
phentsize: "how big is each entry?" → the per-entry size.

Kept phentsize rather than phsize since the latter is ambiguous (per-entry size vs. whole-table size). Let me know if you have any specific comments/recommendations on this one.

core/pta/qcom/pas/pas_auth_core.c
zelvam95: FYI -> In the other places I have renamed it to align/use _len uniformly leaving above one place as exception due to above reasoning.

core/pta/qcom/pas/pas_core.c
zelvam95: These fields are generic MEM_SETUP state: set by pas_core.c, read by the auth path (fw_start/VERIFY_IMAGE reject a zero/stale fw_base). Clearing them here enforces "MEM_SETUP must run again before the next load." The ~10 fw_shutdown handlers do hardware teardown only and read these while running, so pushing the reset in means duplicating it across all of them. Kept here, gated on shutdown success — set/clear stays with the layer that owns the field.

ta/qcom_pas/src/pas_mbn_parser.c
ldts: what is ph? why not just offset, size?

core/pta/qcom/pas/pas_auth_core.c
ldts: I'd prefer to unify - I dont recall OP-TEE making this distinction

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,421 @@
...
+
+struct elf_info {
+	size_t ehsize;
+	size_t phoff;
+	size_t phentsize;
+	size_t phnum;
+	bool is_64;
+};
+
+static TEE_Result parse_elf(const uint8_t *fw, size_t fw_size,
```
ldts, line 95: why not adding this at the top of the file
```
#define PARSE_EHDR(ehdr, _is_64, phdr_type) do {                        \
        info->is_64     = (_is_64);                                     \
        info->ehsize    = (ehdr)->e_ehsize;                             \
        info->phoff     = (ehdr)->e_phoff;                              \
        info->phentsize = (ehdr)->e_phentsize;                          \
        info->phnum     = (ehdr)->e_phnum;                              \
        if (info->ehsize < sizeof(*(ehdr)) ||                           \
            info->phentsize < sizeof(phdr_type))                        \
                return TEE_ERROR_BAD_FORMAT;                            \
} while (0)
```

then just

```
static TEE_Result parse_elf(const uint8_t *fw, size_t fw_size,
                            struct elf_info *info)
{
        const unsigned char *ident = fw;

        if (fw_size < EI_NIDENT)
                return TEE_ERROR_BAD_FORMAT;

        if (ident[EI_MAG0] != ELFMAG0 || ident[EI_MAG1] != ELFMAG1 ||
            ident[EI_MAG2] != ELFMAG2 || ident[EI_MAG3] != ELFMAG3)
                return TEE_ERROR_BAD_FORMAT;

        switch (ident[EI_CLASS]) {
        case ELFCLASS64: {
                const Elf64_Ehdr *ehdr = (const void *)fw;

                if (fw_size < sizeof(*ehdr))
                        return TEE_ERROR_BAD_FORMAT;

                PARSE_EHDR(ehdr, true, Elf64_Phdr);
                break;
        }
        case ELFCLASS32: {
                const Elf32_Ehdr *ehdr = (const void *)fw;

                if (fw_size < sizeof(*ehdr))
                        return TEE_ERROR_BAD_FORMAT;

                PARSE_EHDR(ehdr, false, Elf32_Phdr);
                break;
        }
        default:
                return TEE_ERROR_BAD_FORMAT;
        }

        return TEE_SUCCESS;
}
```

core/pta/qcom/pas/pas_auth_core.c
zelvam95: Done — renamed hash_size → hash_len everywhere (struct pas_auth_core_ctx, struct pas_mbn, pas_mbn_parse(), hash_verify()), so it's uniform with data_len. This also matches tee_hash_createdigest()'s own datalen/digestlen naming, which doesn't distinguish size vs len either. No behavior change, pure rename.

ta/qcom_pas/src/pas_mbn_parser.c
zelvam95: Dropped the ph abbreviation. Renamed to phdr_offset, phdr_entry_len, phdr_count, elf_hdr_len (matches lib/libutee's own ldelf loader, which never uses ph* either — it just spells things out, e.g. offs, num_bytes).

Kept a phdr_ prefix rather than going fully bare, since pas_mbn_locate() also computes a hash-segment offset in the same function (hash_seg_offset/hash_seg_end/hash_seg_len) — a bare offset there would've collided in meaning with the phdr-table offset. So: bare names where there's no ambiguity, a short prefix only where two different offsets/lengths coexist in the same scope.

Applied the same treatment to pas_auth_core.c's struct elf_info (ehsize/phoff/phentsize/phnum → elf_hdr_len/phdr_offset/phdr_entry_len/phdr_count).

core/pta/qcom/pas/pas_auth_core.c
zelvam95: Added PARSE_EHDR per your suggestion, and gave get_phdr() the same treatment with a matching PARSE_PHDR — it had the identical ELFCLASS64/32 duplication but was writing into 5 separate scalar out-params instead of a struct. Added struct phdr_info (mirrors struct elf_info) so both macros are now genuinely parallel: PARSE_EHDR(e_info, ehdr, is_64) / PARSE_PHDR(p_info, phdr), both (out)->field = (src)->field shape. get_phdr() dropped from 7 params to 4.

Both macros + their structs now sit together at the top of the file, right after the MI_PBT_* bit helpers, before any function — no #undef, since none of the other multi-macro files in the qcom driver tree (cmd_db.c, qcom-wdt.c, qfprom_fuse_region.c, etc.) undef their file-scope macros either.

One deliberate departure from your sketch: I kept the elf_hdr_len/phdr_entry_len size validation as an explicit if outside the macro rather than folding the return into it — checkpatch flags a return inside a macro body as WARNING: Macros with flow control statements should be avoided, and there's no precedent for that pattern anywhere else in the tree. So the macro itself is pure field-copy; the validation stays next to the type-specific sizeof(...) it needs.

ta/qcom_pas/include/qcom_pas_priv.h
ldts: so what is the variable invalidating? what does it gate and why does it belong in the pas_md_slot?

core/pta/qcom/pas/platform/platform_pas.h
ldts: lets just return SUCCESS then - when PAS_AUTH is not enabled, we still expect the firmwares to load and run....returning an error here we would expect an error handler to be bypassed an so on.

ta/qcom_pas/src/pas_auth.c
ldts: I still see the same check perform in line 50 and line 58. line58 is not needed if we return TEE_ERROR_OUT_OF_MEMORY in line 57

```
	slot = get_meta_data_slot(s, pas_id);
	if (!slot) {
		for (i = 0; i < PAS_MD_SLOTS; i++) {
			if (!s->slots[i].used) {
				slot = &s->slots[i];
				goto found;
			}
		}
		EMSG("PAS auth: no free md slot (pas_id=%#"PRIx32")", pas_id);
		return TEE_ERROR_OUT_OF_MEMORY;
	}
```

core/pta/qcom/pas/pas_core.c
zelvam95: Hope is acceptable given above reasoning, let me know if any concerns.

ta/qcom_pas/src/pas_auth.c
zelvam95: Tracking it as part of the other comment.

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,373 @@
...
+
+static bool is_hashed(uint32_t p_type, uint32_t p_flags)
+{
+	if (p_type != PT_LOAD)
+		return false;
+
+	return MI_PBT_PAGE_MODE(p_flags) == MI_PBT_NON_PAGED_SEGMENT &&
+	       MI_PBT_SEGMENT_TYPE(p_flags) != MI_PBT_HASH_SEGMENT &&
+	       MI_PBT_ACCESS_TYPE(p_flags) != MI_PBT_NOTUSED_SEGMENT &&
+	       MI_PBT_ACCESS_TYPE(p_flags) != MI_PBT_SHARED_SEGMENT;
```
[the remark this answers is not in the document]
zelvam95: Not applying this — the predicate is intentional. It matches the vendor reference's segment-eligibility check, which includes the boot-hash program-header entry so the boot hash itself is covered by the hash walk. The suggested change would skip that entry and let the boot-hash segment through unverified, so declining.

ta/qcom_pas/src/qcom_pas.c
```diff
@@ -13,17 +15,39 @@
...
+
 	if (pt != exp_pt)
 		return TEE_ERROR_BAD_PARAMETERS;
 
-	/* Firmware authentication - TODO */
+	res = pas_auth_verify_reset(s, pta_session, params[0].value.a, params);
+	if (res)
+		return res;
 
 	return TEE_InvokeTACommand(pta_session, TEE_TIMEOUT_INFINITE,
```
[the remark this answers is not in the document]
zelvam95: The TOCTOU on the metadata pointer is already closed: pas_auth_save_metadata() copies the caller's buffer into TA memory (TEE_Malloc + memcpy into slot->meta_data), and every subsequent step (AUTHENTICATE, AUTH_AND_RESET) operates on that copy, not on the NS-visible buffer.

The firmware payload the DSP actually executes from is a separate TOCTOU that this PR doesn't close. It can be closed from OP-TEE core / TA-side by locking down the firmware carveout with an xPU driver — we don't have that driver yet, so it's tracked as a follow-up incremental improvement rather than something to land in this series.

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,373 @@
...
+		return res;
+	}
+
+	base = reloc_base(ctx->metadata, &e_info);
+
+	for (i = 0; i < e_info.phdr_count; i++) {
+		get_phdr(ctx->metadata, &e_info, i, &p_info);
+
+		if (!is_hashed(p_info.type, p_info.flags) || !p_info.file_len)
+			continue;
```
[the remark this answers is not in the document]
zelvam95: Not applying — the vendor reference doesn't zero pure-BSS (p_filesz == 0) segments from the auth path either; the loaded firmware initializes its own BSS on startup. Matching reference behavior to avoid divergence.

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,373 @@
...
+
+		if (p_info.mem_len < p_info.file_len)
+			return TEE_ERROR_BAD_FORMAT;
+		if (!check_range(offset, p_info.mem_len, ctx->fw_size))
+			return TEE_ERROR_BAD_FORMAT;
+		if (p_info.mem_len > p_info.file_len) {
+			void *zi = ctx->fw + offset + p_info.file_len;
+			size_t zi_len = p_info.mem_len - p_info.file_len;
+
+			memset(zi, 0, zi_len);
```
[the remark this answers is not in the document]
zelvam95: Not applying — the vendor reference walks the program headers once, using the segment-eligibility predicate inline to decide which entries participate in the hash chain. Restructuring to two passes would diverge from that shape without adding verification coverage. Keeping the single-pass form for parity with the reference.

ta/qcom_pas/include/qcom_pas_priv.h
zelvam95: What invalidates it: pas_auth_save_metadata() clears it (and zeroes slot->mbn) whenever new metadata replaces the slot's contents — re-saving forces a fresh authenticate before the next reset.

What it gates: AUTH_AND_RESET. pas_auth_verify_reset() builds the PTA buffer from slot->mbn.hash_table / hash_table_size, which are only populated once pas_auth_authenticate() (parse now, +metadata sig-auth in the follow-on PR) has completed for this slot.

Why in pas_md_slot: the state is per-peripheral (slots keyed by pas_id) and travels with the mbn it guards.

Went with ready over authenticated because with secboot off nothing actually gets authenticated (parse-only) — ready describes the state the reset path consumes and stays honest across both configs.

Personally, I feel hash_table_valid also makes sense for this particular use-case^; 
That said, slot->ready also seems like a valid alternative name to me & I have renamed to slot->ready to close out on this loop based on the suggestions you provided for this as part of another comment.

core/pta/qcom/pas/platform/platform_pas.h
zelvam95: Done — the CFG_QCOM_PAS_AUTH-disabled stub now returns TEE_SUCCESS, so with authentication compiled out the verify step is a pass-through and firmwares load/run normally. This matches the TA-side stubs in pas_auth.h, which already return TEE_SUCCESS in that config.

ta/qcom_pas/src/pas_auth.c
zelvam95: Fixed, updated as per suggestion.

core/pta/qcom/pas/pas_auth_core.c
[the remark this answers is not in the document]
zelvam95: Correcting my earlier reply — split matches the QTEE reference impl:

Range check — applied. image_range() no longer filters on p_filesz, so BSS-only PT_LOADs participate in min-paddr, max(paddr + memsz), carveout-fit, and entry-in-range.
Pure-BSS zeroing — not applying. QTEE's verify loop skips filesz == 0 the same way; the firmware initializes its own BSS.
BSS tail (p_memsz > p_filesz > 0) is already zeroed before hashing.

ta/qcom_pas/src/pas_mbn_parser.c
```diff
@@ -0,0 +1,276 @@
...
+	} else {
+		const Elf32_Ehdr *ehdr = (const void *)md;
+
+		phdr_offset = ehdr->e_phoff;
+		phdr_entry_len = ehdr->e_phentsize;
+		phdr_count = ehdr->e_phnum;
+	}
+
+	if (phdr_count < 2 || !phdr_offset || phdr_entry_len < phdr_entry_min)
+		return TEE_ERROR_BAD_FORMAT;
```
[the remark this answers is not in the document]
zelvam95: Fix applied. QTEE's reference implementation has the same latent inconsistency (its hash_seg_offset computation at pil.c:371 ignores e_phoff too), so this is defense-in-depth beyond the reference. Also tightened e_phentsize handling to strict equality to match QTEE (CHECK(ehdr.e_phentsize == sizeof(ElfXX_Phdr))).

PR #20, opened 2026-07-09
ta/pta: qcom: pas: authenticate PIL firmware signatures and device bindings

zelvam95:
Stacked on PR #19, which added segment-hash verification. Segment hashing
proves an image matches its own digest table, but not that the image was
signed by a trusted party or is permitted on this device. On a fused
secure-boot Lemans device this PR establishes the image's provenance at
INIT_IMAGE, before the REE loads any segment: OEM certificate chain,
signature, and SW/HW device bindings, all anchored to hardware
fuses. Enabled under the same `CFG_QCOM_PAS_AUTH` knob as PR #19.

Multi-root certificates and anti-rollback enforcement are not part of this
series; both are added in a stacked follow-up, PR #35.

Devices whose secure-boot fuse is unblown skip signature authentication and
rely on PR #19's segment-hash verification alone; segment re-hashing at
AUTH_AND_RESET runs on every device regardless.

## Design

Authentication forks on a single fuse read of the secure-boot state and
root-of-trust anchor, and that read **fails closed**: a fuse-PTA error is
treated as secure-boot *enabled*, so a transient glitch can never downgrade a
secure-booted board to hash-only verification.

The crypto runs in the `qcom_pas` user-TA (mbedTLS-backed cert-chain and
signature verification); fuses are reached through a dedicated fuse PTA so the
TA needs no direct driver dependency. UIE-encrypted and QTI-countersigned
images are refused outright — neither decryption nor countersignature
verification is implemented here.

```mermaid
sequenceDiagram
    participant REE as REE (Linux)
    participant TA as qcom_pas TA
    participant FPTA as fuse PTA
    participant PTA as PAS PTA

    Note over REE,PTA: INIT_IMAGE - signature authentication (secure-boot devices)
    REE->>TA: INIT_IMAGE(pas_id, metadata_blob)
    TA->>PTA: PTA_QCOM_PAS_INIT_IMAGE
    PTA-->>TA: TEE_SUCCESS
    TA->>TA: pas_auth_save_metadata()<br/>TEE-private copy keyed by pas_id
    TA->>FPTA: read secure-boot state + root-of-trust anchor
    FPTA-->>TA: secboot_on, anchor<br/>(read error is treated as secboot_on)
    alt secure boot enabled
        TA->>TA: pas_mbn_parse() extracts hash_table
        TA->>FPTA: read device id fuses
        FPTA-->>TA: fuse values
        TA->>TA: verify cert chain vs fused anchor
        TA->>TA: verify signature over OEM-signed region
        TA->>TA: enforce SW/HW bindings
    else secure boot disabled
        TA->>TA: pas_mbn_parse() extracts hash_table<br/>(no signature auth)
    end
    TA-->>REE: TEE_SUCCESS (or TEE_ERROR_SECURITY on any failure)

    Note over REE,PTA: MEM_SETUP + REE segment load (unchanged from PR 19)

    Note over REE,PTA: AUTH_AND_RESET - segment hash verification + release
    REE->>TA: AUTH_AND_RESET(pas_id, fw_base, fw_size)
    TA->>PTA: PTA_QCOM_PAS_VERIFY_IMAGE with metadata and hash_table
    PTA->>PTA: re-hash each PT_LOAD segment vs hash_table entry
    PTA-->>TA: TEE_SUCCESS
    TA->>PTA: PTA_QCOM_PAS_AUTH_AND_RESET
    PTA-->>TA: TEE_SUCCESS peripheral released from reset
    TA-->>REE: TEE_SUCCESS
```

> Any failure — cert-chain break, signature mismatch, or an SW/HW-binding
> mismatch — fails INIT_IMAGE with `TEE_ERROR_SECURITY`;
> the image is never loaded and the peripheral is never released from reset.

## Testing

- Built `PLATFORM=qcom-lemans CFG_QCOM_PAS_AUTH=y CFG_WERROR=y`; checkpatch
  clean across the series.
- On a Lemans board, both fuse configurations:
  - secure-boot on: positive path authenticates and boots; signature mismatch,
    cert-chain break, and SW/HW-binding mismatch are each rejected and the
    peripheral stays in reset;
  - secure-boot off: signature path skipped, segment-hash verification still
    enforced.


ta/qcom_pas/src/pas_sig_auth.c
```diff
@@ -0,0 +1,794 @@
...
+
+	if (meta.anti_rollback < dev_ver) {
+		EMSG("PAS ARB: image version %"PRIu32" < device %"PRIu32,
+		     meta.anti_rollback, dev_ver);
+		return TEE_ERROR_SECURITY;
+	}
+
+	if (meta.anti_rollback > dev_ver)
+		pas_fuse_blow_pil_rollback_version(meta.anti_rollback);
+
```
TonyJH1: The return value here is discarded, therefore any failure is lost. Should we check the return value?

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,419 @@
...
+		if (paddr < base)
+			return TEE_ERROR_BAD_FORMAT;
+		offset = paddr - base;
+
+		if (!range_ok(offset, filesz, ctx->fw_size))
+			return TEE_ERROR_BAD_FORMAT;
+
+		expected = ctx->hash_table + i * ctx->hash_size;
+
+		res = hash_verify(ctx->hash_algo, ctx->fw + offset, filesz,
```
TonyJH1: Check for memsz and filesz and ensure that ZI is cleared as per earlier comments.

ta/qcom_pas/src/pas_fuse.c
```diff
@@ -0,0 +1,246 @@
...
+TEE_Result pas_fuse_blow_pil_rollback_version(uint32_t version)
+{
+	TEE_Param params[TEE_NUM_PARAMS] = { };
+	TEE_Result res = TEE_ERROR_GENERIC;
+	uint32_t pt = 0;
+
+	params[0].value.a = version;
+	pt = TEE_PARAM_TYPES(TEE_PARAM_TYPE_VALUE_INPUT, TEE_PARAM_TYPE_NONE,
+			     TEE_PARAM_TYPE_NONE, TEE_PARAM_TYPE_NONE);
+	res = fuse_pta_invoke(PTA_QCOM_FUSE_BLOW_PIL_ROLLBACK_VERSION, pt,
```
TonyJH1: In the previous design, the fuse blow was deferred. There was an array containing the versions of the images loaded, then a command from the HLOS would be used to trigger the committing of these fuses. I cannot see where this behaviour is modelled here, but I may be missing something

zelvam95: With QTEE, we support both ARB Fuse update after auth + deferred update Tony. As of now, I've included only the auth time update.

This is similar to how we support boot time provisioning using sec.elf but don't support deferred provisioning of sec.elf. Incrementally in future, we can extend this as required. Hope this is fine. Let me know if you have any comments/thoughts on this.

core/pta/qcom/pas/pas_auth_core.c
zelvam95: **Addressed** — same fix as PR#19 thread 10 (this file is introduced by
PR#19). See the PR#19 thread 10 response for the code and comment; nothing
further is added in PR#20.

ta/qcom_pas/src/pas_sig_auth.c
zelvam95: **Addressed.** `check_anti_rollback()` now propagates the fuse-blow error:

```c
if (meta.anti_rollback > dev_ver) {
    res = pas_fuse_blow_pil_rollback_version(meta.anti_rollback);
    if (res) {
        EMSG("PAS ARB: fuse advance to %"PRIu32
             " failed: %#"PRIx32, meta.anti_rollback, res);
        return res;
    }
}
```

A fuse-blow failure fails the whole `AUTH_AND_RESET` (peripheral stays in
reset) rather than the previous silent-drop behavior.

ta/qcom_pas/include/pas_sig_auth.h
```diff
@@ -0,0 +1,66 @@
...
+ * signature, SW_ID/HW binding and anti-rollback. Mirrors the reference PIL
+ * image-authentication flow, which runs this crypto work BEFORE the REE
+ * loads any ELF segment into the carveout - called from
+ * qcom_pas_seg_hash.c's INIT_IMAGE handling, between parsing the hash table
+ * and trusting it.
+ *
+ * When CFG_QCOM_PAS_AUTH is disabled these calls compile to
+ * no-ops (segment hash size defaults to SHA-384, authentication is skipped),
+ * so qcom_pas_seg_hash.c stays free of build-config conditionals.
+ */
```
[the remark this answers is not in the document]
zelvam95: **Addressed.** The `pas_sig_auth.h` file header now describes the actual
call path and doesn't name any specific caller file:

```
 * Signature-authentication backend for the PAS TA: certificate chain,
 * signature, SW_ID/HW binding and anti-rollback. Runs this crypto work
 * BEFORE the REE loads any ELF segment into the carveout, between parsing
 * the hash table and trusting it.
```

The stale `qcom_pas_seg_hash.c` reference is gone across the whole tree
(grep-verified).

ta/qcom_pas/src/pas_sig.c
```diff
@@ -0,0 +1,757 @@
...
+		/*
+		 * MGF1 is fixed at SHA-256: pas_sig_algo_from_leaf() only
+		 * derives the SHA-256 PSS scheme for this TA (matching digest
+		 * and salt length). The signature is exactly the modulus size;
+		 * any bytes past that in the reserved field are padding.
+		 */
+		mbedtls_pk_rsassa_pss_options opts = {
+			.mgf1_hash_id = MBEDTLS_MD_SHA256,
+			.expected_salt_len = salt_len,
+		};
```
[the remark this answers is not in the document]
zelvam95: **Addressed.** `pas_sig_algo_from_leaf()` produces only
`TEE_ALG_RSASSA_PKCS1_PSS_MGF1_SHA256` for RSA-PSS leaves (digest, MGF1 and
salt-length all SHA-256, matching how MBN images are signed). The `SHA-384`
PSS variant is never produced upstream of `pas_sig_verify_signature()` and is
now explicitly caught by the `default:` case of the `switch (sig_algo)`,
which returns `TEE_ERROR_NOT_SUPPORTED` rather than silently running with
an MGF1 mismatch.

A comment above the SHA-256 PSS case spells this out so nobody adds a
SHA-384 case without also fixing `mgf1_hash_id`:

```c
/*
 * RSA-PSS is SHA-256 only. pas_sig_algo_from_leaf()
 * derives exactly this scheme for RSA leaves (digest,
 * MGF1 and salt length all SHA-256). A SHA-384 PSS
 * sig_algo is therefore never produced and is rejected
 * by the default case below rather than verified with
 * a mismatched MGF1.
 */
```

ta/qcom_pas/src/pas_sig.c
```diff
@@ -0,0 +1,757 @@
...
+	mask = GENMASK_32(PAS_MAX_NUM_ROOT_CERTS - 1, 0);
+	non_avail = (revocation_list & mask) | (~activation_list & mask);
+
+	if (__builtin_popcount(non_avail) >= (int)PAS_MAX_NUM_ROOT_CERTS) {
+		if (root_cert_sel == PAS_DEFAULT_SAFE_ROOT_INDEX)
+			return TEE_SUCCESS;
+		EMSG("PAS auth: no usable root cert, sel %"PRIu32,
+		     root_cert_sel);
+		return TEE_ERROR_SECURITY;
+	}
```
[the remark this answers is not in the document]
zelvam95: `qcom_secboot_get_mrc_info()`. They are always the full 4-bit register
value; the register does not shrink when fewer than 4 roots are provisioned.
For any unprovisioned index N (N >= num_roots), the register-level
"activation" bit is 0, which the mask correctly categorizes as unusable —
matching the hardware's semantics. So using `PAS_MAX_NUM_ROOT_CERTS` for
the mask width is intentional and mirrors the register width.

The caller-supplied `num_roots` is enforced at the top of the function
(`root_cert_sel < num_roots`), which is where the "requested index must be
within the provisioned set" bound lives.

ta/qcom_pas/src/pas_auth.c
```diff
@@ -0,0 +1,216 @@
...
+	uint8_t anchor[PTA_QCOM_FUSE_ROOT_OF_TRUST_SIZE] = { };
+	bool secboot_on = false;
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	res = pas_fuse_get_secboot_and_root_anchor(anchor, &secboot_on);
+	if (res)
+		return false;
+
+	return secboot_on;
+}
```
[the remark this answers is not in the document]
zelvam95: *Addressed.** The `secure_boot_provisioned()` helper is gone entirely; the
decision is now a single inline fuse read that **fails closed**. On a fuse-read
error the code treats the device as secure-boot-*on* rather than unprovisioned,
so a fuse-PTA glitch cannot downgrade a secure-booted board to hash-only
authentication:

```c
res = pas_fuse_get_secboot_and_root_anchor(anchor, &secboot_on);
/*
 * Fail closed on a fuse-read failure: treat as secure-boot enabled.
 * Treating it as "not provisioned" would downgrade a secure-booted
 * board to hash-only authentication for the rest of the boot.
 */
if (res)
    secboot_on = true;
```

Ref: `ta/qcom_pas/src/pas_auth.c:108-115` in commit
`ta: qcom_pas: verify image signature and device bindings`.

zelvam95:
[TonyJH1](https://github.com/TonyJH1), [Jorge A. Ramirez-Ortiz (@ldts)](https://github.com/ldts) & [Sumit Garg (@b49020)](https://github.com/b49020), Can you please help to review the PR and share if you have any additional comments? Have addressed all the previous comments including the ones raised by copilot.

ta/qcom_pas/src/pas_fuse.c
zelvam95: P.S. CFG_QFPROM_PROGRAMMING -> We also have this gating config which is off by default and OEMs/users need to turn it on if they want the fuse write to happen. So with this off, ARB fuse updates will not happen, it will be a bypass.

core/drivers/qcom/qfprom/qfprom_secboot.c
```diff
@@ -0,0 +1,257 @@
...
+	if (!drv->corr_base_va)
+		return TEE_ERROR_BAD_STATE;
+
+	va = drv->corr_base_va + (pa - QFPROM_CORR_BASE);
+	*out = io_read32(va) & mask;
+	return TEE_SUCCESS;
+}
+
+TEE_Result qcom_secboot_get_root_of_trust(uint8_t *hash, size_t len)
+{
```
slategrey-vision-pwrf168, line 72: I was wondering if we should consider reading the SENSE registers rather than CORR in this function. Since the roots are not expected to change after a reboot, using the SENSE values can be used. Would it be possible to make that change?

Additionally, for Nord we currently do not have the ability to read the CORR registers, so switching to SENSE would also help with supporting that use case. (-_-;)

core/drivers/qcom/qfprom/lemans/qfprom_target.h
```diff
@@ -17,6 +17,67 @@
 #define QFPROM_CORR_BASE                        0x00784000
 #define QFPROM_SIZE                             0x4000
 
+/*
+ * SECURE_BOOTn secure-control register array holding the per-code-segment
+ * AUTH_EN and USE_SERIAL_NUM bits. Authentication is required when AUTH_EN
+ * is blown; the OEM root-of-trust anchor lives in the PK_HASH_0 fuse. The
+ * APPS code segment is index 1 within the array.
+ */
```
slategrey-vision-pwrf168, line 25: #define SECURITY_CONTROL_BASE           UL(0x00780000)
#define SECURITY_CONTROL_SIZE           UL(0x10000)

This is the definition for the HOYA target. These point to the RAW region of the fuses.
Can we modify these registers to read from the SENSE region instead? (other than ARB, MRC we should be able to use SENSE registers)

ta/qcom_pas/src/pas_sig_auth.c
```diff
@@ -0,0 +1,445 @@
...
+	TEE_Result res = TEE_ERROR_GENERIC;
+	uint32_t version = 0;
+
+	/* MBN v5 hash tables are always SHA-256; v6 selects via fuse. */
+	res = pas_meta_peek_version(slot->meta_data, slot->meta_data_size,
+				    &version);
+	if (res)
+		return res;
+	if (version == PAS_MBN_VERSION_5) {
+		*hash_len = TEE_SHA256_HASH_SIZE;
```
slategrey-vision-pwrf168, line 398: pas_meta_peek_hash_table_algo add this function for getting this in MBN v7 format can we create an abstraction like this instead of hard coding value in common code? I see there is some in MBN v6 can we have common function to handle this in all three cases.

ta/qcom_pas/src/pas_meta.c
```diff
@@ -0,0 +1,318 @@
...
+		meta->soc_vers[i] = pas_mbn_read_u32(m + (META_OFF_SOC_VERS +
+							  i) *
+						     sizeof(uint32_t));
+	for (i = 0; i < META_NUM_SERIAL_NUM; i++)
+		meta->serial_num[i] = pas_mbn_read_u32(m +
+						       (META_OFF_SERIAL_NUM +
+							i) * sizeof(uint32_t));
+	meta->root_cert_sel = pas_mbn_read_u32(m + META_OFF_ROOT_CERT_SEL *
+						   sizeof(uint32_t));
+	meta->anti_rollback = pas_mbn_read_u32(m + META_OFF_ANTI_ROLLBACK *
```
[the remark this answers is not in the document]
ldts: is anti_rollback being used?

ta/qcom_pas/src/pas_auth.c
```diff
@@ -0,0 +1,190 @@
...
+		EMSG("PAS auth: cannot pick hash size: %#"PRIx32, res);
+		return res;
+	}
+
+	res = pas_fuse_get_secboot_and_root_anchor(anchor, &secboot_on);
+	if (res)
+		secboot_on = true;
+
+	if (secboot_on) {
+		res = pas_sig_auth_hash_len(slot, &hash_len);
```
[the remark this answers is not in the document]
ldts: I dont see how...if secboot_on is false, we always apply 384 dont we?

ta/qcom_pas/src/pas_meta.c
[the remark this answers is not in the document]
zelvam95: The parsing part is done here as part of meta-data parsing which is common. 
The ARB field is used during signature verification to reject lower versions as part of a follow up PR which adds support for ARB -> PR: #35.

core/pta/qcom/pas/pas_auth_core.c
```diff
@@ -0,0 +1,403 @@
...
+			 * fetches it outside the CPU's coherency domain, so the
+			 * zeros must reach DDR before reset release.
+			 */
+			dcache_cleaninv_range(zi, zi_len);
+		}
+
+		expected = ctx->hash_table + i * ctx->hash_len;
+
+		res = hash_verify(ctx->hash_algo, ctx->fw + offset,
+				  p_info.file_len, expected, ctx->hash_len);
```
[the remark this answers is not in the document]
ldts: how can this be not in scope and not even a comment or a big warning?? 

PAS authentication means that the firmware the DSP/copro executes it is authenticated. 

If we leave a hole here by which an attacker can replace the firmware after it authenticated it, this means that this feature it is broken.

This needs to be addressed.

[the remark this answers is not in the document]
zelvam95: We copy the meta data to a carveout during the auth (so until the point of auth, we don't depend purely on NS Memory). 

But, yes, post auth attacker can still copy it as you pointed out, but it was discussed previously and planned to be done incrementally & is not in scope of this PR;

We don't have the xPU3 driver etc to do dyn protections.

I can add a comment here explicitly having that as a "to-do", If that'd help? Can maybe plan to pick this up after the PAS Series lands upstream.

core/drivers/qcom/qfprom/lemans/qfprom_target.h
```diff
@@ -17,6 +17,36 @@
 #define QFPROM_CORR_BASE                        0x00784000
 #define QFPROM_SIZE                             0x4000
 
+#define SECURE_BOOT_APPS_ADDR			(SECURITY_CONTROL_BASE + 0x606c)
+#define SECURE_BOOT_AUTH_EN_BMSK		0x20
+#define SECURE_BOOT_USE_SERIAL_NUM_BMSK		0x40
```
harshaldev27: Can I suggest using `BIT()` macro for defining these masks? Makes it more easy to understand which BIT is being referenced here.

core/drivers/qcom/qfprom/lemans/qfprom_target.h
```diff
@@ -17,6 +17,36 @@
...
+
+#define QFPROM_ROOT_OF_TRUST_BYTE_SIZE		48
+
+#define PK_HASH0_SENSE_ADDR			(SECURITY_CONTROL_BASE + 0x6190)
+
+#define OEM_ID_SENSE_ADDR			(SECURITY_CONTROL_BASE + 0x6138)
+#define OEM_ID_BMSK				0xffff0000
```
harshaldev27: You can use `GENMASK_32()` here.

core/drivers/qcom/qfprom/lemans/qfprom_target.h
```diff
@@ -17,6 +17,36 @@
...
+#define OEM_ID_SHFT				16
+#define MODEL_ID_BMSK				0x0000ffff
+#define MODEL_ID_SHFT				0
+#define JTAG_ID_SENSE_ADDR			(SECURITY_CONTROL_BASE + 0x6130)
+#define JTAG_ID_AUTH_BMSK			0x0fffffff
+#define SERIAL_NUM_SENSE_ADDR			(SECURITY_CONTROL_BASE + 0x6134)
+
+#define TCSR_SOC_HW_VERSION_ADDR		0x01FC8000
```
harshaldev27: This address is the same on all Hoya architectures it seems, can you move this and any similar ones to arch_config.h?

core/drivers/qcom/qfprom/qfprom_secboot.c
```diff
@@ -0,0 +1,202 @@
...
+	uint32_t val = 0;
+
+	if (!enabled)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (!drv->raw_base_va)
+		return TEE_ERROR_BAD_STATE;
+
+	val = io_read32(drv->raw_base_va +
+			(SECURE_BOOT_APPS_ADDR - SECURITY_CONTROL_BASE));
```
harshaldev27: Why don't we just `#define SECURE_BOOT_APPS_OFFSET` and add it to `drv->raw_base_va`? Is there any particular reason for always re-calculating the offset by absolute addresses defined in the headers?

core/drivers/qcom/qfprom/qfprom_secboot.c
```diff
@@ -0,0 +1,202 @@
...
+}
+
+static TEE_Result read_sense_reg(paddr_t pa, uint32_t *out)
+{
+	struct qfprom_context *drv = qfprom_get_context();
+
+	if (!drv->raw_base_va)
+		return TEE_ERROR_BAD_STATE;
+
+	*out = io_read32(drv->raw_base_va + (pa - SECURITY_CONTROL_BASE));
```
harshaldev27: If we stop re-calculating the offset here `(pa - SECURITY_CONTROL_BASE)`, you could just take the offset as input instead of `pa` and make this easier to read.

core/drivers/qcom/qfprom/qfprom_secboot.c
```diff
@@ -0,0 +1,202 @@
...
+	res = read_sense_reg(SERIAL_NUM_SENSE_ADDR, &ids->serial_num);
+	if (res)
+		return res;
+
+	return TEE_SUCCESS;
+}
+
+#define SEGMENT_HASH_ROOT_CERT_SEL_MAX	3U
+
+TEE_Result qcom_secboot_get_segment_hash_len(uint32_t root_cert_sel,
```
harshaldev27: Perhaps a bit of documentation explaining what this function is doing with the returned hash_len would be good.

core/drivers/qcom/qfprom/qfprom_secboot.c
```diff
@@ -0,0 +1,202 @@
...
+					     uint32_t *hash_len)
+{
+	if (!hash_len)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if (root_cert_sel > SEGMENT_HASH_ROOT_CERT_SEL_MAX)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+#if SEGMENT_HASH_SELECT_SUPPORTED
```
harshaldev27: How is this config decided?

core/pta/qcom/pas/pas_auth_core.c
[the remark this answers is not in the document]
zelvam95: sMMU based isolation for NS VM Masters (which are OEM Signed) is expected to be done in NS EL2. The protection for TOCTOU must come from the MMU config in the kernel for NS VM Masters.

xPU based isolation is possible only for masters with separate QADID especially those QTI Signed/etc. and we don't support any QTI Signed images PIL bring up in OPTEE.

I will add a comment in code + commit msg to clarify this @ldts. Thanks for pointing this out.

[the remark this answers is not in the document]
zelvam95: Keeping this comment open to track code comments + commit msg update with above dets.

zelvam95:
This PR is being tracked upstream here for further review -> https://github.com/OP-TEE/optee_os/pull/7937. 

Closing this PR in qcom-next which is based on qcom-next version of hash-verify (#19) since the upstream optee version of hash-verify (https://github.com/OP-TEE/optee_os/pull/7885) has been refactored quite a bit post that & it'd make sense to track this PR further in upstream.

ta/qcom_pas/include/pas_meta.h
```diff
@@ -18,4 +18,46 @@ TEE_Result pas_meta_segment_hash_len(const uint8_t *meta_data,
 				     size_t meta_data_size,
 				     uint32_t *hash_len);
 
+TEE_Result pas_meta_get_root_cert_sel(const uint8_t *meta_data,
```
harshaldev27: Can we have some comments here explaining briefly what these functions are doing?

ta/qcom_pas/src/pas_meta.c
```diff
@@ -52,3 +73,241 @@ TEE_Result pas_meta_segment_hash_len(const uint8_t *meta_data,
...
+		memset(copy + off, 0, block_len);
+}
+
+static void zero_field(uint8_t *copy, size_t copy_len, size_t off)
+{
+	if (off + sizeof(uint32_t) <= copy_len)
+		memset(copy + off, 0, sizeof(uint32_t));
+}
+
+TEE_Result pas_meta_signed_copy(const struct pas_mbn *hs,
```
harshaldev27: Name could be improved to `pas_meta_get_signed_region_copy()`

ta/qcom_pas/include/pas_meta.h
```diff
@@ -18,4 +18,46 @@ TEE_Result pas_meta_segment_hash_len(const uint8_t *meta_data,
...
+				      size_t meta_data_size,
+				      uint32_t *root_cert_sel);
+
+TEE_Result pas_meta_verify_preamble(const uint8_t *meta_data,
+				    size_t meta_data_size,
+				    const uint8_t *hash_table,
+				    uint32_t hash_len);
+
+struct pas_meta {
```
harshaldev27: Nit: The commit message also says that this is extracted from the OEM metadata block in the MBN, perhaps calling it `pas_oem_metadata {` makes it more obvious.

ta/qcom_pas/src/pas_meta.c
```diff
@@ -3,10 +3,31 @@
...
+#include <elf64.h>
 #include <pas_mbn_parser_priv.h>
 #include <pas_meta.h>
+#include <string.h>
+#include <string_ext.h>
 #include <tee_internal_api.h>
 #include <utee_defines.h>
+#include <util.h>
+
+#define META_OFF_MAJOR		0
```
harshaldev27: Nit: `OEM_META_OFF_MAJOR?` and so on for below..

ta/qcom_pas/src/pas_auth.c
```diff
@@ -81,7 +85,9 @@ TEE_Result pas_auth_save_metadata(struct qcom_pas_session *s, uint32_t pt,
 TEE_Result pas_auth_authenticate(struct qcom_pas_session *s, uint32_t pas_id)
 {
 	struct pas_md_slot *slot = get_meta_data_slot(s, pas_id);
+	uint8_t anchor[PTA_QCOM_FUSE_ROOT_OF_TRUST_SIZE] = { };
 	TEE_Result res = TEE_ERROR_GENERIC;
+	bool secboot_on = false;
```
harshaldev27: I suggest we should keep `secboot_on = true` here as well. This should only be set to false when: 
1. `pas_fuse_get_secboot_and_root_anchor()` internally updates it to false, and
2. `pas_fuse_get_secboot_and_root_anchor()` returns true.

ta/qcom_pas/src/pas_sig_auth.c
```diff
@@ -0,0 +1,431 @@
...
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result reject_if_encrypted(const struct pas_mbn *hs)
+{
+	if (!hs->uie_encrypted)
+		return TEE_SUCCESS;
+
+	if (!pas_fuse_get_image_encryption_en())
```
harshaldev27: Can you explain this a bit? If image has uie_encrypted enabled, but the fuse says it should be disabled, this function returns success. Shouldn't both be disabled?

ta/qcom_pas/src/pas_sig_auth.c
```diff
@@ -0,0 +1,431 @@
...
+
+	res = pas_fuse_get_eku_enforcement_en(&eku_enforced);
+	if (res) {
+		EMSG("PAS auth: cannot read EKU enforcement fuse: %#"PRIx32,
+		     res);
+		return res;
+	}
+
+	res = pas_sig_verify_cert_chain(hs->oem_certs, hs->oem_certs_size,
+					eku_enforced, 1, 0, &leaf,
```
harshaldev27: Instead of 0 we can pass `SECBOOT_DEFAULT_ROOT_CERT_SEL`?

core/pta/qcom/pas/pas_auth_core.c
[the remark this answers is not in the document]
zelvam95: Comment added as part of upstream PR.

ta/qcom_pas/src/pas_auth.c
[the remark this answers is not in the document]
zelvam95: https://github.com/OP-TEE/optee_os/pull/7937

pas_sig_auth_segment_hash_len -> This function now runs outside the secboot check. Please refer to the upstream PR for latest code. If any clarifications, still needed, let me know & I'd be happy to clarify.

PR #21, opened 2026-07-13
pta: qcom: pas: lemans: add Camera ICP bring-up

michael-jihan-qcom:
The Camera ICP (Imaging Control Processor, PAS ID 33) firmware must be
loaded and authenticated by OP-TEE before CAMX can use the camera
subsystem on Lemans EVK. Without this support the kernel
camera driver fails to bring up the ICP and camera use case is
unavailable.

Add PAS reset ops for the ICP, register the subsystem in the PAS table.

Tested: camera preview use case tested on Lemans EVK; 
ICP firmware loads, authenticates and executes correctly with camera preview
confirmed functional end-to-end.


core/pta/qcom/pas/pta_qcom_pas.c
```diff
@@ -124,6 +124,28 @@ static TEE_Result qcom_pas_auth_and_reset(uint32_t pt,
...
+	 * camera_fw_start() programs the correct SID1 region.
+	 */
+	if (params[0].value.b) {
+		TEE_Result res = TEE_SUCCESS;
+
+		DMSG("auth_and_reset: mem_setup pas_id=%d fw_base=0x%08x%08x fw_size=0x%x",
+		     (int)params[0].value.a, (unsigned)params[1].value.b,
+		     (unsigned)params[1].value.a, (unsigned)params[0].value.b);
+		res = pas_platform_mem_setup(params[0].value.a, params[0].value.b,
+					     params[1].value.a, params[1].value.b);
```
b49020: Don't do this implicit mem_setup in OP-TEE. The client driver in Linux should invoke `prepare_and_auth_reset()` instead.

core/pta/qcom/pas/platform/lemans/camera.c
```diff
@@ -0,0 +1,117 @@
...
+	IMSG("camera: fw_start done");
+	return TEE_SUCCESS;
+}
+
+static TEE_Result camera_fw_shutdown(struct qcom_pas_data *data)
+{
+	vaddr_t base = icp_sys_base(data);
+	uint32_t reg = 0;
+
+	IMSG("camera: fw_shutdown called pas_id=%u", data->pas_id);
```
b49020: Please drop all this runtime logging. You should just keep 1-2 debug prints only which shouldn't be enabled in the release builds.

core/arch/arm/plat-qcom/hoya/lemans/target_config.h
```diff
@@ -114,4 +115,14 @@
 
 #define RPMH_PDC_GPDSP1_BASE		UL(0x0b2d0000)
 #define RPMH_PDC_GPDSP1_SIZE		UL(0x00002000)
+
+/*
```
ldts: remove comment please

core/pta/qcom/pas/platform/lemans/camera.c
```diff
@@ -0,0 +1,117 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ *
+ * OP-TEE PAS reset sequence for the Camera ICP (Imaging Control Processor)
```
ldts: remove comments

core/pta/qcom/pas/platform/lemans/camera.c
```diff
@@ -0,0 +1,117 @@
...
+	return io_pa_or_va(&data->base, data->size);
+}
+
+static TEE_Result camera_fw_start(struct qcom_pas_data *data)
+{
+	vaddr_t base = icp_sys_base(data);
+	uint32_t sid_start = 0;
+	uint32_t sid_end = 0;
+
+	IMSG("camera: fw_start called pas_id=%u fw_base=0x%lx fw_size=0x%zx",
```
ldts: remove imsg

core/pta/qcom/pas/platform/lemans/camera.c
```diff
@@ -0,0 +1,117 @@
...
+
+	io_write32(base + ICP_SYS_CONTROL, 0);
+
+	IMSG("camera: fw_shutdown done");
+	return TEE_SUCCESS;
+}
+
+static TEE_Result camera_fw_set_state(struct qcom_pas_data *data, bool power_on)
+{
+	IMSG("camera: fw_set_state pas_id=%u power_on=%d",
```
ldts: remove IMSG

core/pta/qcom/pas/platform/lemans/camera.c
```diff
@@ -0,0 +1,117 @@
...
+
+#include "camera.h"
+
+/*
+ * icp_sys_base() - return the mapped virtual address of ICP_SYS_REG_BASE.
+ *
+ * TITAN_A_ICP_SYS_REG_BASE = TITAN_SS_BASE + 0x0 (no offset from TITAN_SS).
+ * data->base.pa = TITAN_SS_BASE, data->size = TITAN_SS_BASE_SIZE (0x200000).
+ */
+static vaddr_t icp_sys_base(struct qcom_pas_data *data)
```
ldts: why do we need this static function?

core/pta/qcom/pas/platform/lemans/camera.h
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+#ifndef _CAMERA_H_
+#define _CAMERA_H_
+
+#include "pas_subsys.h"
+
+/*
+ * ICP_SYS register block.
+ *
```
ldts: remove all comments

ldts (changes requested):
Modify the commit message to succinctly explain the reason for the changes - not the how (we can figure out the who from the code).

in the Github PR, please could you explain the methodology you used to test (not just load the firmware but confirm that it is executing correctly)

core/pta/qcom/pas/pta_qcom_pas.c
michael-jihan-qcom: Done, reverted the change

core/arch/arm/plat-qcom/hoya/lemans/target_config.h
michael-jihan-qcom: DOne

core/pta/qcom/pas/platform/lemans/camera.c
michael-jihan-qcom: Done, Please Review

core/pta/qcom/pas/platform/lemans/camera.c
michael-jihan-qcom: Have removed it and followed similar format as iris.c

core/pta/qcom/pas/platform/lemans/camera.h
michael-jihan-qcom: Done, Please review

core/pta/qcom/pas/platform/lemans/camera.c
```diff
@@ -0,0 +1,119 @@
...
+
+#define ICP_SYS_SID1_END_ADDR		0x004c
+
+#define LX7_WFI_POLL_MAX_RETRY		150
+#define LX7_WFI_POLL_INTERVAL_US	2000
+
+static bool camera_is_standbywfi(vaddr_t base)
+{
+	unsigned int retry = 0;
+
```
ldts: replace the loop using `timeout_elapsed` instead of retry/delay

core/pta/qcom/pas/platform/lemans/camera.c
```diff
@@ -0,0 +1,119 @@
...
+	uint32_t sid_end = 0;
+	uint32_t reg = 0;
+
+	base = io_pa_or_va(&data->base, data->size);
+	if (!base)
+		return TEE_ERROR_GENERIC;
+
+	io_write32(base + ICP_SYS_RESET, ICP_SYS_RESET_FUNC_RESET_BIT);
+
+	sid_start = 0;
```
ldts: can drop reinit of sid_start

core/pta/qcom/pas/platform/lemans/camera.c
```diff
@@ -0,0 +1,119 @@
...
+#include <kernel/delay.h>
+#include <stdint.h>
+#include <string.h>
+#include <trace.h>
+#include <util.h>
+
+#include "camera.h"
+
+#define ICP_SYS_RESET			0x0000
+#define ICP_SYS_RESET_FUNC_RESET_BIT	BIT(4)
```
ldts: remove the _BIT for consistency with STANDBYWFI. Then some of the validation checks below might fit on a single line

ldts (changes requested):

core/pta/qcom/pas/platform/lemans/camera.c
michael-jihan-qcom: replaced inline with timeout_elapsed

core/pta/qcom/pas/platform/lemans/camera.c
michael-jihan-qcom: Done, thankyou for the suggestion.

core/pta/qcom/pas/platform/lemans/camera.c
michael-jihan-qcom: done.

core/pta/qcom/pas/platform/pas_data.h
```diff
@@ -18,6 +18,7 @@
 #define PAS_ID_TURING1		30
 #define PAS_ID_GPDSP0		39
 #define PAS_ID_GPDSP1		40
+#define PAS_ID_CAMERA		33
```
b49020: Can you insert this ID in correct sequence?

core/pta/qcom/pas/platform/lemans/camera.c
```diff
@@ -0,0 +1,102 @@
...
+
+	if (power_on) {
+		if (io_read32(base + ICP_SYS_CONTROL) & ICP_SYS_CONTROL_CPU_EN)
+			return TEE_ERROR_BAD_STATE;
+		return camera_fw_start(data);
+	}
+
+	while (!timeout_elapsed(timeout)) {
+		if (io_read32(base + ICP_SYS_STATUS) & ICP_SYS_STATUS_LX7_STANDBYWFI)
+			return camera_fw_shutdown(data);
```
zelvam95, line 92: `camera_fw_set_state()` — a WFI timeout is silently reported as success
```c
while (!timeout_elapsed(timeout)) {
	if (io_read32(base + ICP_SYS_STATUS) & ICP_SYS_STATUS_LX7_STANDBYWFI)
		return camera_fw_shutdown(data);
	udelay(10);
}

DMSG("camera: WFI timeout, forcing halt");
return camera_fw_shutdown(data);
```
On the power-off path, if the LX7 core never reaches the standby-WFI state
within `LX7_WFI_TIMEOUT_US` (300 ms), the function logs "forcing halt" and
calls `camera_fw_shutdown()` anyway — same as the success path — and
`camera_fw_shutdown()` always returns `TEE_SUCCESS`. `pas_platform_
set_remote_state()` (`pas_core.c:88-95`) propagates this return value
straight through to the PTA's caller, so a caller requesting power-off has
no way to distinguish "ICP cleanly reached WFI and shut down" from "ICP
never confirmed WFI; halt was forced anyway." The `DMSG` is debug-level
only (not visible in a release build's default log level), so in practice
this failure mode is invisible end-to-end.

This mirrors an equivalent, already-flagged pattern in the sibling PR for
CDSP (`cdsp_fw_shutdown()`'s unchecked power-on/idle waits) — worth fixing
here for the same reason: a genuinely stuck/misbehaving subsystem should be
distinguishable from a clean shutdown, especially since the comment itself
("forcing halt") signals the author is aware this is a degraded case.

**Suggest:** return `TEE_ERROR_TIMEOUT` (or otherwise signal the forced-halt
case distinctly) instead of falling through to the same `camera_fw_shutdown()`
call and implicit `TEE_SUCCESS` as the clean path.

michael-jihan-qcom: Addressed the comment, thank you for pointing it out, yes this error would have been masked, have added TEE_ERROR_TIMEOUT

core/pta/qcom/pas/platform/pas_data.h
michael-jihan-qcom: Addressed comment

PR #18, opened 2026-06-29
tlmm: qcom: add gpio and pinctrl driver

siva-potu:

PR #23, opened 2026-07-14
drivers: qcom: add TLMM GPIO and pinctrl driver for lemans

siva-potu:
Adds a Qualcomm TLMM GPIO and pinctrl driver (non-DT model) and enables it for lemans.

Supersedes #18 — rebased on qcom-next, comments cleaned up across all files.


core/drivers/qcom/tlmm/gpio.c
b49020, line 1: Rather than calling it a TLMM driver, use the standard pinctrl naming conventions. Follow naming conventions similar to Linux: `drivers/pinctrl/qcom/pinctrl*`

core/arch/arm/plat-qcom/hoya/lemans/tlmm_soc_data.c
b49020, line 1: Rather move this file as a SoC specific file in the drivers directory.

core/drivers/qcom/tlmm/gpio.c
```diff
@@ -0,0 +1,230 @@
...
+}
+
+TEE_Result tlmm_request_pin(unsigned int pin)
+{
+	uint32_t exceptions = 0;
+
+	if (pin >= TLMM_MAX_GPIOS || pin >= tlmm.desc->num_gpios)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	exceptions = cpu_spin_lock_xsave(&tlmm.lock);
```
zelvam95, line 139: Can we use simple mutex here instead of CPU spin locks which are expensive?

core/drivers/qcom/tlmm/gpio.c
```diff
@@ -0,0 +1,230 @@
...
+void tlmm_release_pin(unsigned int pin)
+{
+	uint32_t exceptions = 0;
+
+	if (pin >= TLMM_MAX_GPIOS || pin >= tlmm.desc->num_gpios)
+		return;
+
+	tlmm_restore_lp(&tlmm, pin);
+
+	exceptions = cpu_spin_lock_xsave(&tlmm.lock);
```
zelvam95, line 161: Same comment; Can we use simple mutex instead of this?

core/drivers/qcom/tlmm/gpio.c
```diff
@@ -0,0 +1,230 @@
...
+					  desc->size);
+	if (!tlmm.base) {
+		EMSG("TLMM: failed to map base 0x%"PRIxPA, desc->base);
+		return TEE_ERROR_GENERIC;
+	}
+
+	tlmm.desc = desc;
+	tlmm.gpio_chip.ops = &tlmm_gpio_ops;
+
+	DMSG("TLMM: mapped base 0x%"PRIxPA" -> va 0x%"PRIxVA
```
zelvam95: Same print in 2 different log levels? Can we keep the logging minimal and not overly verbose and only keep logs which will really help during debugging/errors/etc.?

core/drivers/qcom/tlmm/tlmm_private.h
```diff
@@ -0,0 +1,72 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#ifndef __QCOM_TLMM_PRIVATE_H
+#define __QCOM_TLMM_PRIVATE_H
```
zelvam95, line 7: Can we call tlmm_internal maybe instead of tlmm_private? I believe thats the naming used by other existing drivers/vendors in upstream?

core/drivers/qcom/tlmm/pinctrl.c
```diff
@@ -0,0 +1,243 @@
...
+	if (drive_ma <= 2)
+		return 0;
+	if (drive_ma <= 4)
+		return 1;
+	if (drive_ma <= 6)
+		return 2;
+	if (drive_ma <= 8)
+		return 3;
+	if (drive_ma <= 10)
+		return 4;
```
zelvam95: if (drive_ma <= 2)
    return 0;

return MIN(DIV_ROUND_UP(drive_ma - 2, 2), 7U);

Function can be simplified as above?

core/drivers/qcom/tlmm/gpio.c
```diff
@@ -0,0 +1,230 @@
...
+vaddr_t tlmm_pin_reg(const struct tlmm_chip *chip, unsigned int pin,
+		     uint32_t reg_off)
+{
+	return chip->base
+	       + tlmm_tile_offset(chip, pin)
+	       + (vaddr_t)chip->desc->pin_reg_width * pin
+	       + reg_off;
+}
+
+void tlmm_write_cfg(struct tlmm_chip *chip, unsigned int pin,
```
zelvam95: ### 🟡 `core/drivers/qcom/tlmm/gpio.c:57-70` `tlmm_write_cfg()` and `pinctrl.c:47-74` `tlmm_conf_apply()` — hardware read-modify-write on the shared per-pin `TLMM_GPIO_CFG` register is not serialized against concurrent updates on the **same pin**

`tlmm_write_cfg` is a plain read-modify-write with no locking:
```c
void tlmm_write_cfg(struct tlmm_chip *chip, unsigned int pin,
                    uint32_t clear_mask, uint32_t set_mask)
{
    vaddr_t reg = TLMM_GPIO_CFG(chip, pin);
    uint32_t val = io_read32(reg);
    ...
    val = (val & ~clear_mask) | set_mask;
    io_write32(reg, val);
}
```
Callers include:
- `tlmm_set_direction` (gpio.c:147-161) — no lock
- `tlmm_conf_apply` (pinctrl.c:70-74) — **drops the spinlock** at
  pinctrl.c:52 *before* iterating `tlmm_write_cfg`

Per-pin registers don't share a 32-bit word (the SoC descriptor sets
`pin_reg_width = 0x1000`, so each pin has its own dedicated
`TLMM_REG_CFG` at a page-aligned offset — verified via
`tlmm_soc_data.c:15`). So updates to different pins can't race at the
hardware level, and there's no cross-pin RMW hazard.

The remaining hazard is same-pin concurrent updates: e.g. a TA/driver
calling `gpio_ops.set_direction` on pin N while another thread has an
outstanding `tlmm_conf_apply` for a pin group including N. The
ownership bitmap will reject the second one via
`tlmm_pin_is_owned` in `tlmm_set_direction`, but `tlmm_pin_is_owned`
reads `pin_owners` without holding the lock (gpio.c:83-89), so
ownership state is racy against a concurrent `request_pin` / pinctrl
`conf_apply` in flight.

**Suggest:** either take `chip->lock` around the RMW inside
`tlmm_write_cfg` itself (simplest and most defensive), or document a
"caller must hold `chip->lock`" contract on `tlmm_write_cfg` and
ensure `tlmm_conf_apply` retains the lock across the hardware writes,
not just across the ownership grant.

core/drivers/qcom/tlmm/gpio.c
```diff
@@ -0,0 +1,230 @@
...
+		    uint32_t clear_mask, uint32_t set_mask)
+{
+	vaddr_t reg = TLMM_GPIO_CFG(chip, pin);
+	uint32_t val = io_read32(reg);
+
+	clear_mask &= ~TLMM_CFG_PRESERVE_MASK;
+
+	if (chip->desc->has_egpio && (val & TLMM_CFG_EGPIO_PRESENT))
```
zelvam95: ### 🟡 `core/drivers/qcom/tlmm/gpio.c:65-66` `tlmm_write_cfg()` — `TLMM_CFG_EGPIO_ENABLE` is **unconditionally forced on** for any pin where `TLMM_CFG_EGPIO_PRESENT` is set; this is a behavioural change vs the previous CFG value

```c
if (chip->desc->has_egpio && (val & TLMM_CFG_EGPIO_PRESENT))
    set_mask |= TLMM_CFG_EGPIO_ENABLE;
```
Once this driver runs, any pin the SoC labels as "EGPIO-capable" gets
`TLMM_CFG_EGPIO_ENABLE` set on the very next `tlmm_write_cfg()` call —
regardless of what the caller intended. If the pin was previously being
driven in **non**-EGPIO mode (e.g. a plain SPI SE pin muxed via the FUNC
field), enabling EGPIO mode routes the pin to a different peripheral
group (typically EMAC / secondary MDIO / etc. on Qualcomm SoCs). That's a
non-obvious behavioural change that:
- **can not** be undone by clearing `clear_mask` because
  `clear_mask &= ~TLMM_CFG_PRESERVE_MASK` explicitly filters `HIHYS`
  only, not `EGPIO_ENABLE`;
- silently overrides whatever the SPI/UART/etc. driver in PR #33 or
  future users expect by writing FUNC alone.

If the intent is "EGPIO-capable pins default to EGPIO mode when
touched by TLMM," that should be documented and configurable per pin,
not a global "on if PRESENT." If the intent is "preserve whatever
EGPIO_ENABLE bit was already there," this is the wrong pattern — the
existing bit should be included in the clear mask so the caller
controls the final value, or the value should be preserved with a
`set |= val & TLMM_CFG_EGPIO_ENABLE` pattern.

**Suggest:** confirm the intended semantics of `EGPIO_ENABLE` for the
TLMM driver's callers (pinctrl users generally want FUNC-mux, not
EGPIO passthrough). Then either:
1. Preserve the previous value of the bit (`set |= val &
   TLMM_CFG_EGPIO_ENABLE`), or
2. Expose EGPIO enable/disable via the pinconf group struct so the
   caller can choose per group.

core/drivers/qcom/tlmm/pinctrl.c
```diff
@@ -0,0 +1,243 @@
...
+		unsigned int p = pc->pins[i];
+
+		if (chip->pin_owners[p / 32] & BIT(p % 32)) {
+			EMSG("TLMM: pin %u already owned, conf_apply rejected",
+			     p);
+			cpu_spin_unlock_xrestore(&chip->lock, exceptions);
+			return TEE_ERROR_BUSY;
+		}
+	}
+	for (i = 0; i < pc->pin_count; i++) {
```
zelvam95, line 47: ### 🟡 `core/drivers/qcom/tlmm/pinctrl.c:47-77` `tlmm_conf_apply()` — sets `pc->applied = true` **before** the hardware writes; a subsequent failure would leave the state inconsistent

```c
    ...
    cpu_spin_unlock_xrestore(&chip->lock, exceptions);

    pc->applied = true;                                    /* :54 */

    clear = TLMM_CFG_PULL_MASK  | ...
    ...
    for (i = 0; i < pc->pin_count; i++) {
        DMSG(...);
        tlmm_write_cfg(chip, pc->pins[i], clear, set);     /* :73 */
    }
```

`tlmm_write_cfg` currently returns `void` and has no failure path, so
the flag is functionally accurate today. But if any of the register
writes were changed to be error-returning in the future (e.g. adding a
"probe if register access faulted" recovery, or an `io_write32_checked`),
the `applied = true` at line 54 would already commit `conf_free` to
call `tlmm_restore_lp` on pins that never had their hardware programmed.
Move `pc->applied = true` to **after** the hardware-write loop so the
invariant "applied ↔ hardware programmed" is preserved.

Also, note that `pc->applied` is written without holding `chip->lock`,
while `conf_free` reads it also without holding the lock. Not a data
race today (single writer per conf) but worth being explicit that
`applied` is protected by the "no concurrent free while apply is in
flight" contract.

core/drivers/qcom/tlmm/pinctrl.c
```diff
@@ -0,0 +1,243 @@
...
+
+static void tlmm_conf_free(struct pinconf *conf)
+{
+	struct tlmm_pin_conf *pc = conf->priv;
+	struct tlmm_chip *chip = pc->chip;
+	unsigned int i = 0;
+	uint32_t exceptions = 0;
+
+	if (pc->applied) {
+		exceptions = cpu_spin_lock_xsave(&chip->lock);
```
zelvam95, line 93: Also, can't we use regular mutex that's avail in OP-TEE here? Do we really need cpu_spin_lock_xsave/unlock_xrestore here?

core/drivers/qcom/tlmm/gpio.c
```diff
@@ -0,0 +1,230 @@
...
+		EMSG("TLMM: pin %u already owned", pin);
+		return TEE_ERROR_BUSY;
+	}
+	tlmm.pin_owners[pin / 32] |= BIT(pin % 32);
+	cpu_spin_unlock_xrestore(&tlmm.lock, exceptions);
+
+	return TEE_SUCCESS;
+}
+
+void tlmm_release_pin(unsigned int pin)
```
zelvam95, line 151: ### 🟡 `core/drivers/qcom/tlmm/gpio.c:123-135` `tlmm_release_pin()` — calls `tlmm_restore_lp()` (hardware register writes) **before** acquiring `tlmm.lock`

```c
void tlmm_release_pin(unsigned int pin)
{
    ...
    tlmm_restore_lp(&tlmm, pin);                            /* :130 */

    exceptions = cpu_spin_lock_xsave(&tlmm.lock);
    tlmm.pin_owners[pin / 32] &= ~BIT(pin % 32);           /* :133 */
    cpu_spin_unlock_xrestore(&tlmm.lock, exceptions);
}
```
Between `tlmm_restore_lp()` and the ownership clear, another thread
that observes ownership still set can issue a `set_direction` /
`set_value` and its `tlmm_write_cfg` RMW races against the restore's
own RMW on the same pin's CFG register (item 1 above). If the ordering
"restore hardware → release ownership" is intentional (so callers can
still observe ownership while the register is being restored), the
whole sequence should be inside the lock — or better, the ownership
bit should be cleared **first** (blocking new callers) and the hardware
restored after.

Symmetric in `tlmm_conf_free` (pinctrl.c:79-98): the LP restore
happens inside the lock but the `free(conf)` happens outside — fine.

core/drivers/qcom/tlmm/gpio.c
```diff
@@ -0,0 +1,230 @@
...
+
+#include "tlmm_private.h"
+
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, TLMM_BASE, TLMM_BASE_SIZE);
+
+struct tlmm_chip tlmm = { };
+
+static const struct gpio_ops tlmm_gpio_ops;
+
+uint32_t tlmm_tile_offset(const struct tlmm_chip *chip, unsigned int pin)
```
zelvam95, line 26: ### 🔵 `core/drivers/qcom/tlmm/gpio.c:26-46` `tlmm_tile_offset()` — every register access probes all tiles at O(num_tiles) cost; cache the per-pin tile assignment at init

```c
for (t = 0; t < d->num_tiles; t++) {
    id_reg = chip->base + d->tile_offsets[t]
             + (vaddr_t)d->pin_reg_width * pin
             + TLMM_REG_ID_STATUS;
    if (io_read32(id_reg) & TLMM_ID_STATUS_PRESENT)
        return d->tile_offsets[t];
}
```
`tlmm_tile_offset` is called from `tlmm_pin_reg` (line 48-55), which
in turn is called by every `TLMM_GPIO_CFG` / `TLMM_GPIO_IN_OUT` /
`TLMM_GPIO_LP_CFG` macro. On a multi-tile SoC (e.g. up to 5 tiles per
`TLMM_MAX_TILES`), every GPIO op does up to 5 register reads before
the actual read/write. `set_direction` alone is: (5 probes + 1 RMW-read
+ 1 write). At bring-up speed this doesn't matter, but for a driver
that a real SPI or UART may hit thousands of times per second per pin,
this adds up.

Lemans as configured has `num_tiles = 1` so the early return at
gpio.c:32-33 short-circuits — no impact today. But future multi-tile
SoCs will pay for every access. **Suggest:** cache the tile index per
pin at driver init (small array indexed by pin, filled by probing each
`TLMM_REG_ID_STATUS` once).

core/drivers/qcom/tlmm/pinctrl.c
```diff
@@ -0,0 +1,243 @@
...
+			EMSG("TLMM: conf_apply failed for group %u: %#"PRIx32,
+			     i, res);
+			return res;
+		}
+	}
+
+	return TEE_SUCCESS;
+}
+
+void tlmm_free_pin_state(struct pinctrl_state *state)
```
zelvam95, line 244: `tlmm_free_pin_state()` and `tlmm_apply_pin_state()` — no NULL check on the state pointer

Both dereference `state->confs[i]` / `state->conf_count`
unconditionally. A caller with a stashed pointer that failed to be
built (`tlmm_make_pin_state` on error returns via `return res` but sets
`*out_state` only on success — good, so callers with a properly-checked
return value don't hit this) — but defensive `if (!state) return;` is
cheap and prevents future misuse.

core/arch/arm/plat-qcom/hoya/lemans/target.mk
```diff
@@ -1,6 +1,8 @@
 CFG_DRIVERS_CLK ?= y
 CFG_DRIVERS_QCOM_CLK ?= y
 
+CFG_QCOM_TLMM ?= y
```
zelvam95, line 4: — `CFG_QCOM_TLMM ?= y` places the enable in `target.mk`, but the source list for `tlmm_soc_data.c` is added in `core/arch/arm/plat-qcom/sub.mk:10`, not per-target

The registration is:
```make
# plat-qcom/sub.mk
srcs-$(CFG_QCOM_TLMM) += $(QCOM_ARCH_FAMILY)/$(PLATFORM_FLAVOR)/tlmm_soc_data.c
```
This assumes every target that enables `CFG_QCOM_TLMM` has a
`tlmm_soc_data.c` at that path. If a future target enables TLMM
without providing the SoC descriptor, the build fails with an obscure
"file not found" from make. **Suggest:** either move the srcs-y line
into the target's own `sub.mk` (per-target opt-in), or add a `$(call
force,CFG_QCOM_TLMM,n,...)` when the required file is absent, or a
`$(if $(wildcard ...),..., $(error ...))` guard with a clearer message.

core/arch/arm/plat-qcom/hoya/lemans/tlmm_soc_data.c
```diff
@@ -0,0 +1,21 @@
...
+#include <platform_config.h>
+#include <drivers/qcom/tlmm/tlmm.h>
+#include <util.h>
+
+const struct tlmm_desc tlmm_soc_desc = {
+	.base            = TLMM_BASE,
+	.size            = TLMM_BASE_SIZE,
+	.pin_reg_width   = 0x1000,
+	.num_tiles       = 1,
+	.tile_offsets    = { 0x100000, 0, 0, 0, 0 },
```
zelvam95: Lemans is declared with `num_tiles = 1` and `tile_offsets = { 0x100000, 0, 0, 0, 0 }`; the trailing zeros are non-load-bearing but noisy

The `tlmm_tile_offset` code returns `tile_offsets[0]` for `num_tiles
<= 1`, so entries `[1..4]` are never touched. Zeroing them for
readability is fine — but consider using a designated initializer like
`.tile_offsets = { [0] = 0x100000 }` and let the array's implicit
zero-init do the rest.

core/arch/arm/plat-qcom/hoya/lemans/target.mk
siva-potu: moved the 'srcs-y' registration into core/drivers/qcom/tlmm/sub.mk

core/drivers/qcom/tlmm/tlmm_private.h
siva-potu: I couldn't find any _internal references under core/drivers/ and we've used _priv in our own existing driver (qcom/qfprom/qfprom_priv.h).

Keeping it as tlmm_private.h. If you'd prefer consistency with our qfprom driver, I can change it to tlmm_priv.h

core/drivers/qcom/tlmm/gpio.c
siva-potu: The critical section here is limited to a bitmap update and a register read-modify-write. Given the small scope of the protected region, a mutex may introduce unnecessary overhead, and a spinlock seems more appropriate here.

core/drivers/qcom/tlmm/gpio.c
siva-potu: same as [above](https://github.com/qualcomm-linux/optee_os/pull/23/changes/c4b50546e45dfd13bbabf0dd0de195dc55b38217#r3728205380)

core/drivers/qcom/tlmm/pinctrl.c
siva-potu: same as [above](https://github.com/qualcomm-linux/optee_os/pull/23/changes/c4b50546e45dfd13bbabf0dd0de195dc55b38217#r3728205380)

core/drivers/qcom/tlmm/pinctrl.c
siva-potu: Addressed:
tlmm_apply_pin_state() returns TEE_ERROR_BAD_PARAMETERS on NULL and tlmm_free_pin_state() returns early

core/drivers/qcom/tlmm/gpio.c
zelvam95: CPU spin locks in OP-TEE are expensive based on previous discussion with upstream maintainers. Would request to drop to mutex unless there is some real requirement to use spinlock.

core/drivers/qcom/tlmm/gpio.c
zelvam95: Responded to above comment.

core/drivers/qcom/tlmm/tlmm_private.h
zelvam95: Makes sense, given for a qcom driver we've already used that naming convention; We can align to that here as well;

core/drivers/qcom/tlmm/gpio.c
siva-potu: Fixed - the ownership check, LP restore and ownership clear are now all inside tlmm.lock. set_direction/set_value also take the lock now, so both sides of the CFG RMW are serialised.

core/drivers/qcom/tlmm/pinctrl.c
siva-potu: Addressed - applied = true now happens inside the lock alongside the ownership claim that it mirrors.

applied tracks ownership being claimed, not hardware programming completion. It is used by conf_free() to determine whether the pins need to be released. Moving it after the write loop could leave the pins claimed but applied == false on an early-return path, resulting in the ownership never being released. restore_lp() on a pin that was never programmed is a no-op unless LP_CFG_APPLIED is set, so there is no downside to setting applied earlier.

Added a comment in the code to clarify this behavior. The write is now performed under the lock as noted; the read in conf_free() remains unlocked.

core/drivers/qcom/tlmm/gpio.c
siva-potu: sure, will change to mutex

core/drivers/qcom/tlmm/gpio.c
siva-potu: This is an intentional design choice.

The tile probe reads GPIO_ID_STATUS from the pin's register page. TLMM pin registers are XPU-protected at per-pin granularity (pin_reg_width), and not every pin is necessarily accessible from the secure side. Pins may be assigned to HYP or another domain, and probing them can trigger an XPU violation.

I hit exactly this issue in the boot-image TLMM driver, where an eager tile-index cache populated during initialization faulted in UEFI because some pins were HYP-owned and not readable from TZ. As a result, I intentionally avoided an init-time probe of all pins here as well.

With the current approach, we only probe pins that a consumer actually requests, avoiding accesses to unrelated XPU-protected pin pages.

core/arch/arm/plat-qcom/hoya/lemans/tlmm_soc_data.c
siva-potu: addressed

core/drivers/qcom/tlmm/gpio.c
siva-potu: removed the unnecessary logging.

core/drivers/qcom/tlmm/pinctrl.c
siva-potu: updated as per the suggestion

core/drivers/qcom/tlmm/gpio.c
siva-potu: Addressed. The EGPIO handling is now an explicit tlmm_claim_egpio_unlocked() helper instead of being folded into the CFG write's set_mask.

EGPIO_PRESENT is read-only and set by hardware for pads shared with an Island TLMM (LPASS/SSC). EGPIO_ENABLE selects ownership: 1 gives the pad to the chip TLMM, while 0 leaves it with the Island domain. Since the bit resets to 0, shared pads start out Island-owned and CFG/IN_OUT programming has no effect until ownership is claimed. For that reason, ownership is claimed unconditionally for EGPIO_PRESENT pins rather than preserving the current value.

core/drivers/qcom/tlmm/gpio.c
```diff
@@ -0,0 +1,262 @@
...
+	else
+		tlmm_write_cfg_unlocked(tc, pin, TLMM_CFG_OE, 0);
+	mutex_unlock(&tc->lock);
+}
+
+static enum gpio_level tlmm_get_value(struct gpio_chip *chip, unsigned int pin)
+{
+	struct tlmm_chip *tc = chip_to_tlmm(chip);
+
+	assert(pin < tc->desc->num_gpios);
```
ldts: **Functional bug: read-side GPIO ops bounds-check `pin` only via `assert()` (compiled out under `NDEBUG`).**

`tlmm_get_value()` and `tlmm_get_direction()` guard `pin` with `assert(pin < tc->desc->num_gpios)` only. Unlike the write ops (`tlmm_set_value`/`tlmm_set_direction`), which reject out-of-range pins at runtime via `tlmm_pin_is_owned_unlocked()`, and with no `ngpio` check in the GPIO framework, a caller-supplied out-of-range `pin` in a production (`NDEBUG`) build reaches `TLMM_GPIO_IN_OUT(tc, pin)` = `base + pin*0x1000`, an out-of-bounds MMIO read past the `TLMM_BASE_SIZE` mapping. Add the same runtime bounds check the write path uses.

b49020:
Please rebase to tip of qcom-next

core/drivers/qcom/tlmm/gpio.c
siva-potu: sure, will fix this in the rebase commit

core/drivers/qcom/tlmm/gpio.c
```diff
@@ -0,0 +1,270 @@
...
+	mutex_unlock(&tc->lock);
+}
+
+static const struct gpio_ops tlmm_gpio_ops = {
+	.get_direction  = tlmm_get_direction,
+	.set_direction  = tlmm_set_direction,
+	.get_value      = tlmm_get_value,
+	.set_value      = tlmm_set_value,
+};
+DECLARE_KEEP_PAGER(tlmm_gpio_ops);
```
b49020: Pager isn't supported on Qcom SoCs, drop these declarations here and in other places.

b49020:
@siva-potu please create an upstream PR for this TLMM/pinctrl driver since further review can happen there. After addressing one comment I have above we should be good to merge this in `qcom-next`.

core/drivers/qcom/tlmm/gpio.c
siva-potu: Dropped DECLARE_KEEP_PAGER

b49020:
@siva-potu where is the upstream PR link to OP-TEE repo: https://github.com/OP-TEE/optee_os/pulls?

b49020:
I will merge this in `qcom-next` though hoping you create an upstream PR.

b49020:
@siva-potu any updates on the upstream PR?

siva-potu:
i will raise the PR today

siva-potu:
@b49020, raised the upstream PR - https://github.com/OP-TEE/optee_os/pull/8029

PR #25, opened 2026-07-17
Qup clk lemans

nnunna94:
Registers each QUPv3 serial-engine RCG on lemans as a standard struct clk with no parent (clk_ops: enable/disable/set_rate/get_rate), consumed on demand by a future TEE-side SPI/I2C driver via qcom_clk_get_by_name() (this platform has no secure DT, so DT-based acquire isn't available).

set_rate walks a per-domain frequency-config table (mux/divider/MND/DFS-index), and votes a CX/MX voltage corner via RPMh around the rate change — raise before programming, lower after — using an aggregate reference-counted vote model. The voltage vote requires cmd_db_get_addr()/cmd_db_get_aux() to resolve the rail's RPMh resource address and its supported-corner ordinal list, so CFG_QCOM_CMD_DB/CFG_QCOM_RPMH_CLIENT are now force-enabled whenever CFG_QCOM_CLK_BSP=y (default y).

Also renames clock-qcom*.c → clk-qcom*.c for naming consistency with the new clk-framework-based files, and adds cmd_db_get_aux() as a prerequisite API.

Known gaps (documented, not blocking):

The voltage vote is bypassed on a hardware-driven DFS switch (only covers explicit set_rate).
qcom_clk_enable_dfs() isn't serialized against a concurrent clk_set_rate() via the global clk lock.


core/drivers/clk/qcom/sub.mk
```diff
@@ -1,8 +1,10 @@
 global-incdirs-y += .
 global-incdirs-y += platform/$(PLATFORM_FLAVOR)
 
-srcs-y += clock-qcom.c
-srcs-$(CFG_QCOM_PAS_PTA) += platform/$(PLATFORM_FLAVOR)/clock-qcom-pas.c
+srcs-y += clk-qcom.c
+srcs-$(CFG_QCOM_PAS_PTA) += platform/$(PLATFORM_FLAVOR)/clk-qcom-pas.c
+srcs-$(CFG_QCOM_CLK_BSP) += platform/$(PLATFORM_FLAVOR)/clk-qcom-lemans.c
```
zelvam95, line 6: Can we avoid flavor name in the file since we already have PLATFORM_FLAVOR folder and the C files are inside that?

That way we can scale this without updating sub.mk for future targets.

core/arch/arm/plat-qcom/hoya/lemans/target.mk
```diff
@@ -26,3 +26,12 @@ ifeq ($(CFG_QCOM_PAS_PTA),y)
 CFG_RESERVED_VASPACE_SIZE ?= (256 * 1024 * 1024)
 CFG_IN_TREE_EARLY_TAS += qcom_pas/cff7d191-7ca0-4784-af13-48223b9a4fbe
 endif
+
+# QUPv3 serial-engine (bus) clock set-rate/DFS walker, consumed on-demand by a
+# future TEE-side SPI/I2C driver. Set-rate votes CX/MX via RPMh, so pull
+# cmd_db/RPMh client in whenever the walker is built.
+ifeq ($(CFG_QCOM_CLK_BSP),y)
```
zelvam95, line 30: Currently this QUP APIs are not consumed by any drivers I believe?
Ideally its recommended to add at least one consuming client/driver for any of the features we add in upstream so it will be a tested code.
Can you please plan to include either SPI or I2C as part of this Patch Series? (as a separate commit?)

core/drivers/qcom/cmd_db/cmd_db.c
zelvam95, line 1: The commit msg for changes on this file should have 
drivers: qcom: cmd_db: Add cmd_db auxiliary data blob support

Commit msg can then indicate what/why this is required & it'll be consumed by which driver/etc.

The commit title format is followed as above for any drivers I believe; You can check git log core/drivers/qcom

zelvam95: Also, I see, the signed-off-by is missing in the commits; You need to add that. You can refer to existing commits for reference.

Also, if this is AI Assisted you need to add that via Assisted-by tag -> https://www.trustedfirmware.org/aipolicy/

core/drivers/clk/qcom/sub.mk
nnunna94: This makes sense... technically lemans.c is just a bunch of BSP configuration...we can name it clk-qcom-bsp.c, will update the commit.

core/arch/arm/plat-qcom/hoya/lemans/target.mk
nnunna94: i'm working with buses team offline on this... i beleive they started working on spi code... if i can get a smaller version of their PR ..i can try to integrate that into this PR...

core/drivers/qcom/cmd_db/cmd_db.c
nnunna94: I agree this commit message is not really in a standard format... will get this updated. as per standard...including signoff similar to TFA..

core/drivers/clk/qcom/platform/lemans/clk-qcom-lemans.c
```diff
@@ -0,0 +1,207 @@
...
+	{ 100000000, MUX_GPLL0_DIV2, 6,    0,     0, 0x06, CX_SVS },
+	{ 120000000, MUX_GPLL0_DIV2, 5,    0,     0, 0x07, CX_SVS },
+};
+
+/*
+ * 100 MHz-capable QUP SE frequency plan.
+ * Identical to the 120 MHz plan minus its top (120 MHz) row.
+ */
+static const struct qcom_clk_mux_config qup_se_100mhz[] = {
+	{   7372800, MUX_GPLL0_DIV2, 2,  384, 15625, QCOM_CLK_DFS_IDX_NONE, CX_MIN_SVS },
```
zelvam95, line 60: Should we use some macros for these numbers?
I see these numbers are re-used across structs. If we can come up with macros, it'd help avoid any typos/etc.?

core/drivers/clk/qcom/sub.mk
zelvam95: Sure, also, do we really have to call it as clk-qcom-bsp.c?
If its really qup specific, should we rather call it clk-qcom-qup.c?

BSP seems way too generic? (Both for config & file name?)

core/drivers/clk/qcom/platform/lemans/clk-qcom-lemans.c
nnunna94: these are like dividers that needs to programmed into a mux div registers to achieve a given frequency...these can vary as new frequencies pop up for each target. maintaining macros might become tedious and overkill here.

core/drivers/qcom/cmd_db/cmd_db.c
zelvam95: https://optee.readthedocs.io/en/latest/general/contribute.html#commit-messages

You can refer to the optee read the doc for OPTEE's commit msgs format/expectations.

core/drivers/clk/qcom/sub.mk
nnunna94: technically this file can be extended to any other clock configuration in future... like smmu in downstream or for that instance any single clock that needs to be enabled can be placed here and called via clk enable ... I agree today its just for qup but bsp would be a better name i beleive.

core/drivers/clk/qcom/platform/lemans/clk-qcom-lemans.c
zelvam95: The first field appears to be the frequency. I was initially considering a named constant for readability, but Linux uses the raw frequency value here as well, so let's leave it as-is.

core/drivers/clk/qcom/sub.mk
zelvam95: Ideally, we want the respective clients to call clk_enable directly from their driver like how its done for QFPROM today. 
Not sure we fore-see another separate clk-bsp driver need for any specific clocks.
For pas, we needed this, and we already have a separate file for that.
We don't have SMMU support/plan to have SMMU config in OP-TEE.

Maybe, @b49020 / @ldts can also share their feedback on this one. We can plan based on that;

PR #26, opened 2026-07-22
clk: qcom: lemans: add QUP SE clock set-rate/DFS/enable support

nnunna94:
This series adds QUPv3 serial-engine (QUP SE) clock support to the Qualcomm
lemans target, for on-demand use by a TEE-side SPI/I2C driver.


core/drivers/clk/qcom/platform/lemans/clk-qcom-bsp.c
```diff
@@ -0,0 +1,207 @@
...
+		      GCC_CLOCK_BRANCH_ENA_VOTE_1, 13),
+	QUP_SE_DOMAIN("gcc_qupv3_wrap0_s4_clk", 0x13624, qup_se_100mhz,
+		      GCC_CLOCK_BRANCH_ENA_VOTE_1, 14),
+	QUP_SE_DOMAIN("gcc_qupv3_wrap0_s5_clk", 0x13758, qup_se_100mhz,
+		      GCC_CLOCK_BRANCH_ENA_VOTE_1, 15),
+	QUP_SE_DOMAIN("gcc_qupv3_wrap0_s6_clk", 0x1388c, qup_se_100mhz,
+		      GCC_CLOCK_BRANCH_ENA_VOTE_1, 16),
+
+	/*
+	 * QUPv3 wrapper 1: S0-S1 at 120 MHz, S2-S6 at 100 MHz. S0-S5 VOTE_1,
```
zelvam95, line 173: no runtime protection for the domain explicitly flagged as "must never be touched"
```c
/*
 * QUPv3 wrapper 1: S0-S1 at 120 MHz, S2-S6 at 100 MHz. S0-S5 VOTE_1,
 * S6 VOTE_3. S3 is the console UART's QUP SE and must never be
 * enabled/disabled/set-rate by this driver or its self-test.
 */
...
QUP_SE_DOMAIN("gcc_qupv3_wrap1_s3_clk", 0x144f0, qup_se_100mhz,
	      GCC_CLOCK_BRANCH_ENA_VOTE_1, 25),
```
`gcc_qupv3_wrap1_s3_clk` is registered as an ordinary domain in the table
like every other SE, with no special-case handling anywhere in
`clk-qcom.c`'s `qcom_qup_clk_enable()`/`disable()`/`set_rate()`,
`qcom_clk_get_by_name()`, or `qcom_clk_enable_dfs()`. The only thing
preventing a future consumer (the commit message names "a future TEE-side
SPI/I2C driver") from calling `qcom_clk_get_by_name("gcc_qupv3_wrap1_s3_clk")`
and then disabling/reconfiguring it — which per the comment would silently
kill the OP-TEE console — is the comment itself.

**Suggest:** either omit this domain from the table entirely (if nothing
should ever drive it through this API) or add an explicit guard in
`qcom_clk_get_by_name()`/the `clk_ops` callbacks that rejects operations on
it, so the "must never" constraint is enforced rather than only documented.

core/drivers/clk/qcom/sub.mk
```diff
@@ -1,8 +1,10 @@
 global-incdirs-y += .
 global-incdirs-y += platform/$(PLATFORM_FLAVOR)
 
-srcs-y += clock-qcom.c
-srcs-$(CFG_QCOM_PAS_PTA) += platform/$(PLATFORM_FLAVOR)/clock-qcom-pas.c
+srcs-y += clk-qcom.c
+srcs-$(CFG_QCOM_PAS_PTA) += platform/$(PLATFORM_FLAVOR)/clk-qcom-pas.c
+srcs-$(CFG_QCOM_CLK_BSP) += platform/$(PLATFORM_FLAVOR)/clk-qcom-lemans.c
```
zelvam95: `CFG_QCOM_CLK_BSP` hardcodes `clk-qcom-lemans.c` instead of `clk-qcom-$(PLATFORM_FLAVOR).c`
```make
srcs-y += clk-qcom.c
srcs-$(CFG_QCOM_PAS_PTA) += platform/$(PLATFORM_FLAVOR)/clk-qcom-pas.c
srcs-$(CFG_QCOM_CLK_BSP) += platform/$(PLATFORM_FLAVOR)/clk-qcom-lemans.c
```
The `CFG_QCOM_PAS_PTA` line correctly parameterizes on `$(PLATFORM_FLAVOR)`
for both the source path and the filename. The new `CFG_QCOM_CLK_BSP` line
one row below parameterizes the *directory* on `$(PLATFORM_FLAVOR)` but
hardcodes the *filename* to `clk-qcom-lemans.c`. Today this is harmless
because only `lemans/target.mk` sets `CFG_QCOM_CLK_BSP ?= y` (confirmed:
`kodiak` never sets it, and there's no global default anywhere in
`mk/config.mk` or `core/drivers/clk/qcom/sub.mk` that would flip it on for
another flavor) — but the moment a second platform enables
`CFG_QCOM_CLK_BSP=y`, the build will look for
`platform/<that-flavor>/clk-qcom-lemans.c`, which won't exist, and fail.

**Suggest:** rename `clk-qcom-lemans.c` to `clk-qcom-$(PLATFORM_FLAVOR).c`'s
expansion for lemans (i.e. keep today's lemans filename as-is only if that
happens to already match the intended convention) or change the `sub.mk`
line to `platform/$(PLATFORM_FLAVOR)/clk-qcom-$(PLATFORM_FLAVOR).c`,
matching the line above it.

core/drivers/clk/qcom/platform/lemans/clk-qcom-bsp.c
nnunna94: I think this comment section popped up when i added a self unit test and it went ahead and modified the se rate... the client driver is aware of the se that is used for uart logging and is programmed already in xbl ideally. here in optee, bus driver just enables DFS for the same se, nothing more i believe as per buses driver requirement. i should remove those comments to avoid confusion. buses driver is the only consumer for all these domains across images and they are aware as to which domain scale in which mode @zelvam95.

core/drivers/clk/qcom/sub.mk
nnunna94: There was a discussion on this file naming on last PR, we discussed on keeping it as clk_bsp.c and you suggest qup.c i beleive. since then i did not rename it... will rename it.

core/drivers/clk/qcom/platform/lemans/clk-qcom-bsp.c
zelvam95: Yes, please clean up all comments that don't necessary add value/any qTEE references/etc.. and we can retain only comments that don't explain the obvious/adds value to upstream community/maintainers/etc.

core/drivers/clk/qcom/sub.mk
zelvam95: Yes I remember that discussion. Feel free to rename that if that makes sense. 

FYI -> The above comment is slightly different & is not particularly pointing at that. This is related to how we have added the below src include in sub.mk. I want to avoid the explict mention of lemans in the RHS for the file name since that would require you to modify the sub.mk for each platform/cause platform when you try to enable this for some platform other than lemans.

`srcs-$(CFG_QCOM_CLK_BSP) += platform/$(PLATFORM_FLAVOR)/clk-qcom-**lemans**.c`

For ex: If you enable this for lets say Kodiak, it will search for the file platform/kodiak/clk-qcom-lemans.c instead of clk-qcom-kodiak.c. Hope this particular comment is clear now.

ldts:
please rework all the abstractions and propose again:. you have a added a new driver (a voltage regulator) and hidden it under clock. doesnt seem like something we want to do.

ldts:
rename -bsp.c to -cfg.c

nnunna94:
> please rework all the abstractions and propose again:. you have a added a new driver (a voltage regulator?) and hidden it under clock. doesnt seem like something we want to do.
> 
> if it is just a clock helper but unrelated to the actual clock, what about just qcom_corner_vote.c, or qcom_rail_vote.c?

Thats just a file to aggregate on voltage requests from clock driver side and send it to RPMH driver.  this voting mechanism changes based on architecture so had to separate it out from clk-qcom.c to support both shikra and lemans with same clk-qcom.c the file name may be adjusted accordingly.

ldts:
> > please rework all the abstractions and propose again:. you have a added a new driver (a voltage regulator?) and hidden it under clock. doesnt seem like something we want to do.
> > if it is just a clock helper but unrelated to the actual clock, what about just qcom_corner_vote.c, or qcom_rail_vote.c?
> 
> Thats just a file to aggregate on voltage requests from clock driver side and send it to RPMH driver. this voting mechanism changes based on architecture so had to separate it out from clk-qcom.c to support both shikra and lemans with same clk-qcom.c the file name may be adjusted accordingly

so what is the abstraction then? is this common to architecture or platform?  do you need to move clk/qcom/platform to clk/qcom/$arch/$platform and then have clk/qcom/hoya/qcom_corer_vote.c ?

nnunna94:
> > > please rework all the abstractions and propose again:. you have a added a new driver (a voltage regulator?) and hidden it under clock. doesnt seem like something we want to do.
> > > if it is just a clock helper but unrelated to the actual clock, what about just qcom_corner_vote.c, or qcom_rail_vote.c?
> > 
> > 
> > Thats just a file to aggregate on voltage requests from clock driver side and send it to RPMH driver. this voting mechanism changes based on architecture so had to separate it out from clk-qcom.c to support both shikra and lemans with same clk-qcom.c the file name may be adjusted accordingly
> 
> so what is the abstraction then? is this common to architecture or platform? do you need to move clk/qcom/platform to clk/qcom/$arch/$platform and then have clk/qcom/hoya/qcom_corer_vote.c ?

The idea was to use this file for all hoya and  wildcat architectures, for any rpm based architectures introduce a qcom_vreg_rpm.c file or so and based on chipset we decide which file to include in mk files keeping the API interfaces same for clk-qcom.c to remain the same. As of now we did not introduce any architecture folders as such.

ldts:
> The idea was to use this file for all hoya and wildcat architectures, for any rpm based architectures introduce a qcom_vreg_rpm.c file or so and based on chipset we decide which file to include in mk files keeping the API interfaces same for clk-qcom.c to remain the same. As of now we did not introduce any architecture folders as such.

ok so it is not per architecture then. so lets have  qcom_vreg_rpmh.c and qcom_vreg.h?  the problem with vreg is that one expects a regulator driver...not sure if we can come with a better name equally short..

nnunna94:
> > The idea was to use this file for all hoya and wildcat architectures, for any rpm based architectures introduce a qcom_vreg_rpm.c file or so and based on chipset we decide which file to include in mk files keeping the API interfaces same for clk-qcom.c to remain the same. As of now we did not introduce any architecture folders as such.
> 
> ok so it is not per architecture then. so lets have qcom_vreg_rpmh.c and qcom_vreg.h? the problem with vreg is that one expects a regulator driver...not sure if we can come with a better name equally short..

fortunately, hoya and wildcat has this same RPMH interface so file is common for both these architectures. SHIKRA is RPM based so it uses different interface that power drivers team provides.  the file name is something we can keep that all of us can aggree i guess that should not be a problem. i can understand the vreg naming misleading to voltage regulator.

ldts:
ok so so lets make the backend name explicit in the same directory (ie qcom_vreg_rpmh.c ? ). And then in the future, when we need another backend, we can the create a vreg/ and move all backends there.

ldts (changes requested):
change as discussed - clarification of corner voltages.

IDineshChoudhary (changes requested):
RPMH and CMD_DB drivers already exist, can you check the paths mentioned

ldts (changes requested):
Remove AI slop from commit messages. But please do it globally or we will never get this merged. 
For example, why do you need to add this to the commit message?
```
While here, make copy_aux_data() return TEE_ERROR_SHORT_BUFFER instead
of silently truncating an undersized read (unreached by either
existing caller, which both query metadata only), and clear
result->len on the zero-length-entry path cmd_db_get_aux() exercises.
```

Also Remove unnecessary AI  comments from the code (those comments were written by AI to explain _you_ the logic it was using: unless you consider them really necessary - something that you wouldn't be able to figure out-  you don't need to commit those.

Do not create a new file in a commit and then rename it on the subsequent commit in the same series: just fix it on the first commit.

Do not add me as a co-author: I am just reviewing and providing suggestions.

core/drivers/qcom/cmd_db/cmd_db.c
```diff
@@ -372,6 +383,32 @@ TEE_Result cmd_db_get_addr(const char *res_id, uint32_t *addr)
...
+	result.len = *len;
+
+	mutex_lock(&query_db.lock);
+
+	if (!query_db.data) {
+		mutex_unlock(&query_db.lock);
+		return TEE_ERROR_BAD_STATE;
+	}
+
+	res = cmd_db_get_entry_by_res_id(res_id, &result, buf);
```
IDineshChoudhary: this i/f API is static within cmd_db.c, if it's exported than we don't need additional wrapper?
@zelvam95 can you check new i/f for rpmh & cmd_db.

zelvam95: Hi Dinesh,

When I pushed Fuse Mgmt for Hoya, I was supposed to only make those APIs required by Fuse drivers as public. That is the only rationale for current design.
 
If we don't have any concerns with clients calling them directly, we could definitely make them non-static / public as/when required.

zelvam95: On that note -> In TF-A also I believe cmd_db_get_aux is a separate public API exposed to clients (In Sumit's series IIRC). So, if it'd make sense to align with TF-A, we can align here.

In general If multiple cmd_db_get_aux clients will be there, then it'd make sense to expose it as separate API like how we've done now along with those additional checks etc. (to avoid duplicate checks in each client/calling place) but if we will have only one or so caller, then we won't need wrapper and we can directly expose the cmd_db_get_entry_by_res_id by making it as non-stack and adding to header (basically we can try to expose the required abstractions to clients and ensure the code overall is concise/clean/doesn't have redundant checks)

nnunna94:
> > on the the second comment related to conditional compilation on .c files, currently kodiak or other targets do not have the DTPM use cases so they dont need these APIs. we have a strict rule not to use conditional compilation inside .c files, then i would need to either go with a new file and make it compiled for lemans or wherever need via .mk files or i end up compiling this this new functions regardless of any conditions... just that these new APIs remain as dead code for kodiak or targets wherever they dont need them. what would be our take on this ? let me know so that i can address the second query accordingly.
> 
> the pattern we use everywhere is the abstraction in the header file (conditional compilation), and the C file conditionally compiled under config.
> 
> example:
> 
> * https://github.com/qualcomm-linux/optee_os/blob/qcom-next/core/arch/arm/plat-qcom/diag_log.h#L14
> * https://github.com/qualcomm-linux/optee_os/blob/qcom-next/core/arch/arm/plat-qcom/sub.mk#L5

@ldts / @zelvam95 i have addressed the rpmh_vote.h file naming issue, the comments generated by AI issue, the commit messages clean ups to most extent. the one thing left to address is the #if def inside .c file changes. the problem i am seeing here with the header file approach is that some of my functions inside the ifdef are static ones not declared in any header files, these are local to the .c file. The header file solution doesnt really solve the problem for here i believe since those are not public or declared in header fiels to add stubs in header files. Also making them non static is also not really gonna solve the problem as well i believe. one solution is to move all this new functions and APIs to a new file outside of clk-qcom.c and make things align to standards. for now, i am implementing this, let me know if you see any concern here.i borrowed the name for that file from linux usptream.. since most of the operations in that file revolve around rcg configuration.... Also, the one thing left out... the cmd_db related commit, i would need to work it out with rpmh team @IDineshChoudhary further if i can get a PR for this and the fix identified on the MSG ID length issue or should we refactor it here on the same PR further.

pradeep-kondoju:
looks good to me from soc infra.

core/drivers/clk/qcom/sub.mk
```diff
@@ -1,8 +1,18 @@
...
-srcs-y += clock-qcom.c
+srcs-y += clk-qcom.c
+srcs-$(CFG_QCOM_CLK_CFG) += clk-qcom-rcg.c
 srcs-$(CFG_QCOM_PAS_PTA) += platform/$(PLATFORM_FLAVOR)/clock-qcom-pas.c
+srcs-$(CFG_QCOM_CLK_CFG) += platform/$(PLATFORM_FLAVOR)/clk-qcom-cfg.c
+
+# Aggregate CX/MX rail-corner voting behind the set-rate path. RPMh targets
+# build the RPMh backend; without it a CLK_CFG build fails to link rather than
+# silently running under-volted.
+ifeq ($(CFG_QCOM_CLK_CFG),y)
```
ldts: why not just add 

```suggestion
ifeq ($(CFG_QCOM_CLK_CFG),y)
srcs-y += platform/$(PLATFORM_FLAVOR)/clk-qcom-cfg.c
srcs-$(CFG_QCOM_RPMH_CLIENT) += rpmh_rail.c
srcs-y += clk-qcom-rcg.c
endif
```

nnunna94: updated

vinod-amanaganti (changes requested):
please address the review comments.

core/drivers/clk/qcom/clk_ops.c
```diff
@@ -0,0 +1,1000 @@
...
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	for (i = 0; i < *nb_elts; i++)
+		rates[i] = desc->configs[start_index + i].freq_hz;
+
+	return TEE_SUCCESS;
+}
+
+static const struct clk_ops rcg_ops = {
```
ldts: **Rail vote is never released when the clock is disabled.**

`rcg_ops` has no `.disable` hook, and the CX/MX rail vote is only lowered inside `rcg_set_rate()` when a *lower* rate is subsequently requested (`next_corner < curr_corner`). So once an RCG is set to a high rate, `priv->corner` keeps the rail asserted even after the branch is gated off — and it stays up until someone calls `set_rate` with a lower corner, which a disabled consumer typically will not do.

For the on-demand SPI/I2C consumer this targets (enable → transact → disable), the SVS-corner vote on CX/MX leaks for the lifetime of the system after the first transaction.

Is this intentional (rate persists in hardware across gating, so the rail must stay up)? If so, please add a comment documenting the choice. Otherwise the rail vote should be dropped on disable and re-asserted on enable so each rail is only held while some consumer actually needs it — which also matches the per-corner refcounting rationale in the `rail_vote` commit.

ldts: with the above addressed

`Reviewed-by: Jorge Ramirez-Ortiz <jorge.ramirez@oss.qualcomm.com>`

core/drivers/clk/qcom/clk_ops.c
```diff
@@ -0,0 +1,1000 @@
...
+		return TEE_ERROR_ITEM_NOT_FOUND;
+
+	if (!find_src_index(priv->desc, cfg->src, &new_idx, &mux_sel))
+		return TEE_ERROR_ITEM_NOT_FOUND;
+
+	new_parent = clk_get_parent_by_index(clk, new_idx);
+	if (!new_parent)
+		return TEE_ERROR_ITEM_NOT_FOUND;
+
+	old_parent = clk_get_parent(clk);
```
vinod-amanaganti: can you mark it as curr_parent for better readability

core/drivers/clk/qcom/clk_ops.c
nnunna94: addressed this with proper disable ops.

core/drivers/clk/qcom/clk_ops.c
nnunna94: will update this in upcoming iterations @vinod-amanaganti

core/drivers/clk/qcom/clk_ops.c
```diff
@@ -0,0 +1,1021 @@
...
+static void rcg_switch_parent_post(struct clk *clk, struct clk *old_parent,
+					struct clk *new_parent)
+{
+	if (old_parent)
+		pll_unref(old_parent);
+
+	clk->parent = new_parent;
+}
+
+static TEE_Result rcg_set_rate(struct clk *clk, unsigned long rate,
```
ldts: Could you please investigate whether this is a valid concern: unlike `rcg_enable()` and `rcg_disable()` (which both early-return on `priv->dfs`), `rcg_set_rate()` has no `priv->dfs` guard. If a consumer calls `clk_set_rate()` after `qcom_clk_enable_dfs()` has enabled hardware DFS on this RCG, `config_mux_regs()` + `rcg_commit_config()` would write CFG/CFG_UPDATE while the hardware performance-state machine owns the RCG (DFS_EN set), racing it. If that call ordering is possible for this clock, a `priv->dfs` guard here (as in enable/disable) may be warranted.

core/drivers/clk/qcom/clk_ops.c
```diff
@@ -0,0 +1,1021 @@
...
+		const struct clk_mux_config *c = &desc->configs[i];
+
+		if (c->dfs_idx == CLK_DFS_NA)
+			continue;
+
+		if (seen++ < start_index)
+			continue;
+
+		rates[filled] = c->freq_hz;
+		dfs_indices[filled] = c->dfs_idx;
```
ldts: Could you please investigate whether this is a valid concern: `qcom_clk_get_dfs_rates_array()` validates `!count` and handles `!rates` (the size-query path), but never checks `dfs_indices` before writing `dfs_indices[filled]` here. A caller that passes a non-NULL `rates` buffer but a NULL `dfs_indices` would hit a NULL dereference. If `dfs_indices` is meant to be optional, a guard is needed; if it's mandatory, rejecting NULL up front (alongside the other param checks) would make the contract explicit.

vinod-amanaganti:
Address above two comments, rest looks good.
Reviewed-by:  vinod-amanaganti <vinoda@qti.qualcomm.com>

ldts:
> Address above two comments, rest looks good.
> Reviewed-by: vinod-amanaganti [vinoda@qti.qualcomm.com](mailto:vinoda@qti.qualcomm.com)

addressed how? I saw no change in the code and no comments to the request of the investigation.

core/drivers/clk/qcom/clk_ops.c
nnunna94: addressed

core/drivers/clk/qcom/clk_ops.c
nnunna94: addressed

vinod-amanaganti:
Please update the copyrights.

nnunna94:
> Please update the copyrights.

updated copyrights

ldts:
please run scripts/checkpatch.sh on each commit and fix the warnings

nnunna94:
Ran Checkpatch on the commits and updated the latest commit series.  the only warnings left are something recommending to use u32 instead of uint32. resolved most of them.

ldts:
You can add my reviewed-by:

`Reviewed-by: Jorge Ramirez-Ortiz <jorge.ramirez@oss.qualcomm.com>`

b49020:
@ldts feel free to merge this since you have mostly reviewed this PR. I will let @nnunna94 post this series upstream where I can help further review it.

nnunna94:
> You can add my reviewed-by:
> 
> `Reviewed-by: Jorge Ramirez-Ortiz <jorge.ramirez@oss.qualcomm.com>`

is this something i need to add in commit messages ?

zelvam95:
> > You can add my reviewed-by:
> > `Reviewed-by: Jorge Ramirez-Ortiz <jorge.ramirez@oss.qualcomm.com>`
> 
> is this something i need to add in commit messages ?

Yes right, please refer to the commit msgs I shared as reference. You can have my tag as well -

`Reviewed-by: Selvam Sathappan Periakaruppan <speriaka@qti.qualcomm.com>`

nnunna94:
> > > You can add my reviewed-by:
> > > `Reviewed-by: Jorge Ramirez-Ortiz <jorge.ramirez@oss.qualcomm.com>`
> > 
> > 
> > is this something i need to add in commit messages ?
> 
> Yes right, please refer to the commit msgs I shared as reference. You can have my tag as well -
> 
> `Reviewed-by: Selvam Sathappan Periakaruppan <speriaka@qti.qualcomm.com>`

added Reviewed-by Tags to all 4 commits.

zelvam95:
> > > > You can add my reviewed-by:
> > > > `Reviewed-by: Jorge Ramirez-Ortiz <jorge.ramirez@oss.qualcomm.com>`
> > > 
> > > 
> > > is this something i need to add in commit messages ?
> > 
> > 
> > Yes right, please refer to the commit msgs I shared as reference. You can have my tag as well -
> > `Reviewed-by: Selvam Sathappan Periakaruppan <speriaka@qti.qualcomm.com>`
> 
> added Reviewed-by Tags to all 4 commits.

@nnunna94, Have you raised upstream PR for this? Can you please share it here?

I think for merging in qcom-next, raising an upstream PR is like a gating check now (expected to be enforced). We could request Jorge to merge it once you have raised the upstream PR.

nnunna94:
> > > > > You can add my reviewed-by:
> > > > > `Reviewed-by: Jorge Ramirez-Ortiz <jorge.ramirez@oss.qualcomm.com>`
> > > > 
> > > > 
> > > > is this something i need to add in commit messages ?
> > > 
> > > 
> > > Yes right, please refer to the commit msgs I shared as reference. You can have my tag as well -
> > > `Reviewed-by: Selvam Sathappan Periakaruppan <speriaka@qti.qualcomm.com>`
> > 
> > 
> > added Reviewed-by Tags to all 4 commits.
> 
> nnunna94, Have you raised upstream PR for this? Can you please share it here?
> 
> I think for merging in qcom-next, raising an upstream PR is like a gating check now (expected to be enforced). We could request Jorge to merge it once you have raised the upstream PR.

https://github.com/OP-TEE/optee_os/pull/7996 Upstream PR for the same

PR #27, opened 2026-07-22
drivers: qcom: add OP-TEE-owned XPU protection for TZDRAM and DIAG log

thariaruchamy-bit:
## Summary

Move XPU protection of TZDRAM and the OP-TEE DIAG log from TF-A's static
policy into OP-TEE, which already owns the values being protected. Callers
only describe the region and access policy; the driver resolves the
guarding XPU instance and a free resource group on its own.

## Testing

Verified by inspecting the XPU resource-group registers programmed by
OP-TEE (addresses, permissions, enable/lock bits) and by triggering XPU
violations from non-secure accesses to both regions.

Tested-on: Hermosa (IPQ52xx)
Tested-on: Juhu (IPQ96xx)

### DIFF
--- a/core/drivers/qcom/xpu/xpuv4.c
+++ b/core/drivers/qcom/xpu/xpuv4.c
@@ -1,0 +1,255 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#include <assert.h>
+#include <drivers/qcom/xpuv4.h>
+#include <initcall.h>
+#include <io.h>
+#include <kernel/tee_misc.h>
+#include <mm/core_memprot.h>
+#include <mm/core_mmu.h>
+#include <platform_config.h>
+#include <tee_api_types.h>
+#include <trace.h>
+#include <types_ext.h>
+#include <util.h>
+#include <xpu_target.h>
+
+/*
+ * XPU4 register offsets.
+ * IDR0 reports the highest resource group (RG) index in NRG. Each RG
+ * occupies 0x40 bytes starting at base + 0x1000.
+ *   IDR0     +0x0000  NRG = bits 25:16
+ *   RGCR1n   +0x1004  bit 0 = RGE (enable)
+ *   RGCSAR1n +0x1008  start address [63:32]
+ *   RGCSAR0n +0x100C  start address [31:0]
+ *   RGCEAR1n +0x1010  end   address [63:32]
+ *   RGCEAR0n +0x1014  end   address [31:0]
+ *   RGRDRn   +0x1018  read  QAD vector
+ *   RGWRRn   +0x101C  write QAD vector
+ *   QADRGLn  +0x1030  QAD lock vector (bit 31 = secure lock RGL_S)
+ */
+#define XPU4_IDR0(b)        ((b) + 0x0u)
+#define XPU4_RGCR1n(b, n)   ((b) + 0x1004u + 0x40u * (n))
+#define XPU4_RGCSAR1n(b, n) ((b) + 0x1008u + 0x40u * (n))
+#define XPU4_RGCSAR0n(b, n) ((b) + 0x100Cu + 0x40u * (n))
+#define XPU4_RGCEAR1n(b, n) ((b) + 0x1010u + 0x40u * (n))
+#define XPU4_RGCEAR0n(b, n) ((b) + 0x1014u + 0x40u * (n))
+#define XPU4_RGRDRn(b, n)   ((b) + 0x1018u + 0x40u * (n))
+#define XPU4_RGWRRn(b, n)   ((b) + 0x101Cu + 0x40u * (n))
+#define XPU4_QADRGLn(b, n)  ((b) + 0x1030u + 0x40u * (n))
+
+#define XPU4_IDR0_NRG_MASK	SHIFT_U32(0x3FF, 16)
+#define XPU4_IDR0_NRG_SHIFT	16
+
+#define XPU4_RGCR1n_RGE		BIT32(0)
+
+/*
+ * struct xpu_instance - An XPU4 instance and the memory it guards
+ * @base:        XPU4 register base (SoC physical address)
+ * @map_size:    Size of the register window to map at @base
+ * @addr_off:    Base the instance's address registers are relative to
+ * @guard_start: Start of the memory range this instance guards
+ * @guard_size:  Size of the memory range this instance guards
+ * @rg_start:    First resource group available for dynamic allocation
+ * @rg_count:    Number of resource groups available from @rg_start
+ *
+ * Resource groups outside [@rg_start, @rg_start + @rg_count) belong to
+ * other owners such as earlier boot stages and are never allocated here.
+ */
+struct xpu_instance {
+	paddr_t base;
+	size_t map_size;
+	paddr_t addr_off;
+	paddr_t guard_start;
+	size_t guard_size;
+	unsigned int rg_start;
+	unsigned int rg_count;
+};
+
+static const struct xpu_instance xpu_instances[] = {
+	{
+		.base = DDR_MPU_BASE,
+		.map_size = CORE_MMU_PGDIR_SIZE,
+		.addr_off = DDR_MPU_ADDR_OFFSET,
+		.guard_start = DRAM0_BASE,
+		.guard_size = DRAM0_SIZE,
+		.rg_start = DDR_XPU_RG_START,
+		.rg_count = DDR_XPU_RG_COUNT,
+	},
+	{
+		.base = IMEM_MPU_BASE,
+		.map_size = IMEM_MPU_MAP_SIZE,
+		.addr_off = IMEM_BASE,
+		.guard_start = IMEM_BASE,
+		.guard_size = IMEM_SIZE,
+		.rg_start = IMEM_XPU_RG_START,
+		.rg_count = IMEM_XPU_RG_COUNT,
+	},
+};
+
+register_phys_mem_pgdir(MEM_AREA_IO_SEC, DDR_MPU_BASE, CORE_MMU_PGDIR_SIZE);
+register_phys_mem(MEM_AREA_IO_SEC, IMEM_MPU_BASE, IMEM_MPU_MAP_SIZE);
+
+static const struct xpu_instance *xpu_find_instance(paddr_t start, size_t size)
+{
+	const struct xpu_instance *inst = NULL;
+	size_t i = 0;
+
+	for (i = 0; i < ARRAY_SIZE(xpu_instances); i++) {
+		inst = &xpu_instances[i];
+
+		if (core_is_buffer_inside(start, size, inst->guard_start,
+					  inst->guard_size))
+			return inst;
+	}
+
+	return NULL;
+}
+
+static TEE_Result xpu_find_free_rg(vaddr_t base,
+				   const struct xpu_instance *inst,
+				   unsigned int *rg)
+{
+	unsigned int rg_end = inst->rg_start + inst->rg_count;
+	unsigned int nrg = 0;
+	unsigned int i = 0;
+
+	/* IDR0 reports the highest RG index, so the count is one more */
+	nrg = ((io_read32(XPU4_IDR0(base)) & XPU4_IDR0_NRG_MASK) >>
+	       XPU4_IDR0_NRG_SHIFT) + 1;
+	if (rg_end > nrg) {
+		EMSG("XPU: RG window %u..%u exceeds %u RGs implemented",
+		     inst->rg_start, rg_end - 1, nrg);
+		return TEE_ERROR_BAD_STATE;
+	}
+
+	for (i = inst->rg_start; i < rg_end; i++) {
+		if (io_read32(XPU4_RGCR1n(base, i)) & XPU4_RGCR1n_RGE)
+			continue;
+
+		*rg = i;
+		return TEE_SUCCESS;
+	}
+
+	EMSG("XPU: no free RG in window %u..%u", inst->rg_start, rg_end - 1);
+	return TEE_ERROR_OUT_OF_MEMORY;
+}
+
+static void xpu4_program_rg(vaddr_t base, unsigned int rg, paddr_t start,
+			    paddr_t end, uint32_t read_qad, uint32_t write_qad,
+			    uint32_t lock_qad)
+{
+	io_write32(XPU4_RGCSAR1n(base, rg), (uint32_t)(start >> 32));
+	io_write32(XPU4_RGCSAR0n(base, rg), (uint32_t)start);
+	io_write32(XPU4_RGCEAR1n(base, rg), (uint32_t)(end >> 32));
+	io_write32(XPU4_RGCEAR0n(base, rg), (uint32_t)end);
+	io_write32(XPU4_RGRDRn(base, rg), read_qad);
+	io_write32(XPU4_RGWRRn(base, rg), write_qad);
+	io_write32(XPU4_RGCR1n(base, rg), XPU4_RGCR1n_RGE);
+	io_write32(XPU4_QADRGLn(base, rg), lock_qad);
+}
+
+/*
+ * xpu_protect_region() - Restrict access to a memory region
+ * @start:      Region start address (must be 4KB aligned)
+ * @size:       Region size in bytes (must be 4KB aligned, non-zero)
+ * @read_perm:  Read QAD permission vector
+ * @write_perm: Write QAD permission vector
+ *
+ * Restricts [@start, @start + @size) to @read_perm/@write_perm and locks
+ * the configuration. The XPU instance guarding the region and a free
+ * resource group within it are determined from @start, so callers only
+ * describe the region and the access policy they need.
+ *
+ * The lock is held by whichever AP domains (secure/non-secure APPS) are
+ * present in @read_perm/@write_perm.
+ */
+static TEE_Result xpu_protect_region(paddr_t start, size_t size,
+				     uint32_t read_perm, uint32_t write_perm)
+{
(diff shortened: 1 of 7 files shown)


core/drivers/qcom/xpu/ac_xpuv4.c
```diff
@@ -0,0 +1,110 @@
...
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	xpu4_program_rg(base, rg, xpu_start, xpu_start + size,
+			read_perm, write_perm, lock_perm);
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result ac_xpu_init(void)
```
zelvam95: Instead of ac_xpu_init, should we give some other meaningful name, given here what we do is ideally TZDRAM (OPTEE DDR) protection?

core/arch/arm/plat-qcom/diag_log.c
```diff
@@ -78,6 +79,16 @@ void qcom_diag_log_init(void)
 {
 	struct diag *diag = NULL;
 	uint32_t *diag_info_addr = NULL;
+#if defined(CFG_QCOM_XPUV4)
```
zelvam95: Should we split this into separate commit? Given its like a client/user of the driver below?

Can we avoid the if defined check and call that ac_xpu_protect_region by default and have stubs for that ac_xpu_protect_region function for targets that don't have this XPU config enabled?

core/drivers/qcom/xpu/ac_xpuv4.c
thariaruchamy-bit: Renamed as ddr_xpu_init

core/drivers/qcom/xpu/xpuv4.c
```diff
@@ -0,0 +1,109 @@
...
+
+	xpu4_program_rg(base, rg, xpu_start, xpu_start + size,
+			read_perm, write_perm, lock_perm);
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result ddr_xpu_init(void)
+{
+	/* Compile-time guarantee for our known TZDRAM constants */
```
zelvam95: Can we remove the obvious comments from this file?

core/arch/arm/plat-qcom/diag_log.c
```diff
@@ -13,13 +14,26 @@
 
 #include "diag_log.h"
 
+#ifndef IMEM_MPU_BASE
+#define IMEM_MPU_BASE		0
```
zelvam95: What exactly is the goal here? If its not defined Can we assume 0? Won't that cause issues? Should we rather throw error to define it/force the platforms to define it?

core/drivers/qcom/xpu/xpuv4.c
```diff
@@ -0,0 +1,109 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ *
+ * XPU access-control protection for OP-TEE secure memory regions.
```
zelvam95: We can drop this comment;

core/drivers/qcom/xpu/xpuv4.c
```diff
@@ -0,0 +1,109 @@
...
+	io_write32(XPU4_RGCEAR1n(base, rg), (uint32_t)(end >> 32));
+	io_write32(XPU4_RGCEAR0n(base, rg), (uint32_t)end);
+	io_write32(XPU4_RGRDRn(base, rg), read_qad);
+	io_write32(XPU4_RGWRRn(base, rg), write_qad);
+	io_write32(XPU4_RGCR1n(base, rg), 1);  /* RGE=1: enable the RG */
+	io_write32(XPU4_QADRGLn(base, rg), lock_qad);  /* lock config */
+}
+
+TEE_Result xpu_protect_region(paddr_t xpu_base, size_t map_size,
+			      unsigned int rg, paddr_t addr_off,
```
somegupt: RG allocation should be handled by the XPU driver rather than individual teams. The driver should select an available RG at runtime and configure it accordingly.

core/drivers/qcom/xpu/xpuv4.c
```diff
@@ -0,0 +1,109 @@
...
+	io_write32(XPU4_RGCSAR0n(base, rg), (uint32_t)start);
+	io_write32(XPU4_RGCEAR1n(base, rg), (uint32_t)(end >> 32));
+	io_write32(XPU4_RGCEAR0n(base, rg), (uint32_t)end);
+	io_write32(XPU4_RGRDRn(base, rg), read_qad);
+	io_write32(XPU4_RGWRRn(base, rg), write_qad);
+	io_write32(XPU4_RGCR1n(base, rg), 1);  /* RGE=1: enable the RG */
+	io_write32(XPU4_QADRGLn(base, rg), lock_qad);  /* lock config */
+}
+
+TEE_Result xpu_protect_region(paddr_t xpu_base, size_t map_size,
```
somegupt: xpu_base should be identified at run-time based the region that needs protection, should not be passed as an argument from the caller.

core/drivers/qcom/xpu/xpuv4.c
```diff
@@ -0,0 +1,206 @@
...
+	/* The XPU stores addresses relative to its memory-region base */
+	xpu_start = start - inst->addr_off;
+
+	xpu4_program_rg(base, rg, xpu_start, xpu_start + size,
+			read_perm, write_perm, lock_perm);
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result ddr_xpu_init(void)
```
zelvam95: `ddr_xpu_init()` — commit message's stated motivation doesn't match what the code does
The commit message says:
> The final TZDRAM base and size are not fixed. They depend on the memory
> profile selected at runtime... and that layout is known only to OP-TEE.

But `ddr_xpu_init()` protects `CFG_TZDRAM_START`/`CFG_TZDRAM_SIZE`, which
are plain Makefile-level, build-time-fixed constants
(`core/arch/arm/plat-qcom/bobcat/ipq52xx/target.mk:6-7`:
`CFG_TZDRAM_START ?= 0x87D80000`, `CFG_TZDRAM_SIZE ?= 0x280000`; same
pattern for `ipq96xx/target.mk:9-10`). There is no runtime memory-profile
selection mechanism anywhere in `plat-qcom` today (searched for
`mem_profile`/`MEMORY_PROFILE`/`premium`/`low_mem` — no hits). As written,
`ddr_xpu_init()` reads exactly the same static value TF-A's own build would
use for the same platform, just resolved from OP-TEE's Makefile instead of
TF-A's — which is a legitimate reason to move ownership (single source of
truth, no risk of the two firmware images disagreeing), but it is a
different justification than "OP-TEE knows something at runtime that TF-A
can't know at build time." Either the runtime memory-profile mechanism
exists elsewhere and isn't visible in this diff (in which case a
code comment or commit-message pointer to it would help), or the commit
message should be corrected to describe the actual benefit (build-time
single-source-of-truth, not runtime knowledge).

Above was AI Agent's feedback. 
On that note, something to think about -> we don't support LM Profile for any QCOM targets still. 
Should we maybe add that also as a single commit as part of this PR for ipq52xx_lm (the motivation behind this xPU in OPTEE)? Or we can at least update clearly that same targets will have such lm profiles added in future for which this is required to ensure OPTEE can protect its own stuff without depending on TF-A.

core/arch/arm/plat-qcom/diag_log.c
```diff
@@ -78,6 +79,14 @@ void qcom_diag_log_init(void)
 {
 	struct diag *diag = NULL;
 	uint32_t *diag_info_addr = NULL;
+#if defined(CFG_QCOM_XPUV4)
```
zelvam95: `qcom_diag_log_init()` — runs (and calls `xpu_protect_region()`) twice on any build with the project's default `CFG_CORE_ASLR=y`, wasting scarce resource groups
`qcom_diag_log_init()` is invoked via `plat_trace_init()` →
`console_init()`. `console_init()` is called from **two** places in
`core/arch/arm/kernel/entry_a64.S`:
- Line 382, immediately after cache invalidation and **before**
  `core_init_mmu_map` (line 429) / `enable_mmu` (line 453) — i.e. pre-MMU.
- Line 492, inside an `#ifdef CFG_CORE_ASLR` block, explicitly commented
  "Reinitialize console, since register_serial_console() has previously
  registered a PA and with ASLR the VA is different from the PA" — this
  second call happens **after** `enable_mmu` (line 453).

`CFG_CORE_ASLR` defaults to `y` project-wide (`mk/config.mk:346`) and is
not overridden to `n` anywhere in `plat-qcom`, so **on the project's
default configuration, `console_init()` — and therefore
`qcom_diag_log_init()` and this PR's new `xpu_protect_region()` call —
runs twice** on every boot.

Consequence: the second call to `xpu_protect_region(DIAG_BASE, DIAG_SIZE,
...)` finds the RG programmed by the first call still marked `RGE`
(enabled) in `xpu_find_free_rg()` (`xpuv4.c:125-131`, the loop `continue`s
past any RG with `RGE` set), so it allocates and locks a **second,
separate** resource group for the exact same DIAG log region instead of
either being a no-op or returning an error. On `ipq52xx` (Hermosa), whose
entire free OCIMEM-MPU window is only 2 resource groups
(`IMEM_XPU_RG_START = 9`, `IMEM_XPU_RG_COUNT = 2`,
`ipq52xx/target_config.h:34-35`), this single logical protection request
exhausts the *entire* free window, leaving zero resource groups available
for any future OCIMEM-guarded protection need on that platform. On
`ipq96xx` (Juhu, 4 free groups) it consumes half the window for one
region.

This doesn't create a *security* gap — the region ends up protected either
way, redundantly — but it silently burns a small, hardware-fixed resource
that other future OCIMEM protection users on the same platform will need.
It's also worth noting: the commit's own test-verification note says "The
DIAG log was allocated RG10 on Hermosa..., the first group free" — but
`IMEM_XPU_RG_START` for `ipq52xx` is `9`, not `10`. That's consistent with
what this analysis predicts (RG 9 silently consumed by the first, pre-MMU
call; RG 10 is what the developer's later register inspection actually
caught from the second, post-MMU call) — worth confirming against the
actual test log rather than assuming it's just a typo.

**Fix:** guard `qcom_diag_log_init()` (or specifically the
`xpu_protect_region()` call within it) against being run more than once —
e.g. a static "already initialized" flag, or move the XPU call to a
`driver_init()`-ordered initcall (matching how `ddr_xpu_init()` in Commit 1
is wired) instead of running it inline from the console-init path, which is
demonstrably called twice by design.

core/arch/arm/plat-qcom/bobcat/ipq52xx/target_config.h
```diff
@@ -35,4 +35,6 @@
 #define IMEM_XPU_RG_START		UL(9)
 #define IMEM_XPU_RG_COUNT		UL(2)
 
+#define DIAG_XPU_READ_QAD		UL(0xC000000B)
```
zelvam95: Instead of hardcoding QAD value like this which makes it tough to understand what we want to achieve. Can we have individual QAD values defined in the xPU driver related headers and re-use that over here? That way users can explicitly understand which masters have access?

core/arch/arm/plat-qcom/bobcat/ipq96xx/target_config.h
```diff
@@ -32,4 +32,6 @@
 #define IMEM_XPU_RG_START		UL(6)
 #define IMEM_XPU_RG_COUNT		UL(4)
 
+#define DIAG_XPU_READ_QAD		UL(0xC000007B)
```
zelvam95: Same comment as in ipq52xx.

core/arch/arm/plat-qcom/bobcat/ipq52xx/target_config.h
```diff
@@ -25,4 +25,14 @@
...
 
+/* MEMNOC_SCH_MPU: resource groups 10..15 are free for runtime allocation */
+#define DDR_MPU_BASE			UL(0xAC0000)
+#define DDR_MPU_ADDR_OFFSET		UL(0x80000000)
+#define DDR_XPU_RG_START		UL(10)
+#define DDR_XPU_RG_COUNT		UL(6)
+
+/* OCIMEM_MPU: resource groups 9..10 are free for runtime allocation */
+#define IMEM_XPU_RG_START		UL(9)
+#define IMEM_XPU_RG_COUNT		UL(2)
```
b49020: Don't add all these xPU MPUs to the platform configuration files. Instead have then in the drivers directly like: `core/include/drivers/qcom/xpu/bobcat/xpu-ipq52xx.h` and so on.

core/arch/arm/plat-qcom/diag_log.c
```diff
@@ -78,6 +79,31 @@ void qcom_diag_log_init(void)
...
+	 * xpu_protect_region() is safe to call before the MMU is enabled: it
+	 * falls back to a physical address MMIO access in that case.
+	 */
+	static bool xpu_protected;
+
+	if (!xpu_protected) {
+		TEE_Result res = TEE_SUCCESS;
+
+		res = xpu_protect_region(DIAG_BASE, DIAG_SIZE,
+					 DIAG_XPU_READ_QAD, QAD_APPS_SEC);
```
b49020: Rather than calling xPU protect here, why don't you do it in xPU driver itself as part of init sequence sitting alongside TZDRAM? You can surely check if `DIAG_BASE` and `DIAG_XPU_READ_QAD` are defined.

core/drivers/qcom/xpu/xpuv4.c
```diff
@@ -0,0 +1,212 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
vedswarup, line 1: XPU drivers are not needed/planned in optee and is not as per the long term Access control design we are working on. Any XPU settings here if it is dependent on some OEM configurations can be managed in TFA through transfer list or boot params. Please remove xpu related changes from here.

zelvam95: Hi Ved,

Thanks for your response.

From a functional perspective, boot parameters are not a viable option today as there is currently no standard mechanism for TF-A to obtain the complete DDR reservation information associated with OP-TEE. Today, U-Boot SPL only passes the image information (ELF load region/image size) via shared_params, which is insufficient for xPU protection as it does not describe the complete OP-TEE DDR/IMEM footprint, including all reserved memory regions that must be protected.

While Transfer Lists could potentially address the boot-time use case, they would not help with runtime scenarios such as Secure PIL, which are also planned to be supported from OP-TEE on this architecture and these targets. For those runtime use cases, OP-TEE must still have direct ownership and awareness of the memory regions being protected, making the proposed placement less suitable for our current design.

If there are any functional concerns with the driver implementation posted, or if you have specific feedback on the design itself, please let us know and we would be happy to review and address them. However, with respect to driver placement, we would prefer to align with the approach already commercialised on the Bobcat architecture. Moving the functionality at this stage would require significant design and implementation rework across multiple features. It would also introduce a dependency between two boot components that OEMs have historically preferred to avoid and have been quite vocal about keeping independent.

The implementation will remain configuration based, allowing flexibility for future platforms. For a future Bobcat target, we can certainly explore opportunities to converge with the Wildcat approach if it provides clear benefits. For the current set of targets, however, we are planning to proceed with the existing Bobcat design.

Thanks, 
Selvam

core/arch/arm/plat-qcom/bobcat/ipq52xx/target_config.h
thariaruchamy-bit: Sure Sumit, Used a per-flavor directory + fixed filename in the latest change, mirroring what qfprom and clk/qcom already do

core/drivers/qcom/xpu/xpuv4.c
```diff
@@ -0,0 +1,220 @@
...
+				  QAD_APPS_SEC, QAD_APPS_SEC);
+}
+driver_init(ddr_xpu_init);
+
+static TEE_Result diag_xpu_init(void)
+{
+	return xpu_protect_region(DIAG_BASE, DIAG_SIZE,
+				  DIAG_XPU_READ_QAD, QAD_APPS_SEC);
+}
+driver_init(diag_xpu_init);
```
b49020: I would rather call it a `service_init()` sequence. And you can club the 2 xPU protect calls under single service init sequence.

core/include/drivers/qcom/xpuv4.h
```diff
@@ -0,0 +1,60 @@
...
+ * describe the region and the access policy they need.
+ *
+ * The lock is held by whichever AP domains (secure/non-secure APPS) are
+ * present in @read_perm/@write_perm.
+ *
+ * Only declared when CFG_QCOM_XPUV4 is enabled: every caller of this API
+ * lives in code that is itself built only under CFG_QCOM_XPUV4.
+ */
+TEE_Result xpu_protect_region(paddr_t start, size_t size,
+			      uint32_t read_perm, uint32_t write_perm);
```
b49020: Is this API really need to be exposed now since the service using it exists in the driver only?

core/drivers/qcom/xpu/xpuv4.c
```diff
@@ -0,0 +1,220 @@
...
+
+	return xpu_protect_region(CFG_TZDRAM_START, CFG_TZDRAM_SIZE,
+				  QAD_APPS_SEC, QAD_APPS_SEC);
+}
+driver_init(ddr_xpu_init);
+
+static TEE_Result diag_xpu_init(void)
+{
+	return xpu_protect_region(DIAG_BASE, DIAG_SIZE,
+				  DIAG_XPU_READ_QAD, QAD_APPS_SEC);
```
b49020: For DIAG xPU protections, let's add a runtime check if `CFG_QCOM_DIAG_LOG` is enabled.

PR #29, opened 2026-07-27
pta: qcom: ice: add software-key ICE PTA

thariaruchamy-bit:
## Summary

Adds a Qualcomm Inline Crypto Engine (ICE) pseudo-TA that lets the Linux
kernel's dm-crypt/inline-crypt path program raw software keys into ICE key
slots for inline storage encryption, and enables it on the QCOM IPQ (bobcat)
platforms.

Two commands are exposed:
- `PTA_CMD_ICE_INVALIDATE_KEY` — wipe a key slot with random data.
- `PTA_CMD_ICE_SET_CONFIG_KEY` — program key/salt, cipher mode and data-unit
  size (AES-XTS-128/256, AES-CBC-128/256).

Only the REE kernel may open a session on this PTA (`TEE_LOGIN_REE_KERNEL`),
since it is driven by the kernel's inline-crypt engine and not by userspace.

## Testing

Verified on target: booted to the kernel, confirmed an ICE software-key
configuration request is dispatched from the kernel to this PTA, then wrote
and read back a file on the encrypted filesystem and confirmed the plaintext
and decrypted-data md5sums match.

Tested-on: IPQ52xx, IPQ96xx


core/arch/arm/plat-qcom/bobcat/ipq52xx/target_config.h
```diff
@@ -25,4 +25,8 @@
 #define QCOM_WDT_TMR_BASE		UL(0x0B117000)
 #define QCOM_WDT_BARK_INT_ID		UL(0x23)
 
+/* SDC1 SDCC ICE (Inline Crypto Engine) LUT keys register region */
+#define SDCC_ICE_LUT_KEYS		UL(0x0780A000)
```
harshaldev27: Are these addresses same across the Bobcat architecture? If so, I think they should be in arch_config.h

lib/libutee/include/pta_emmc_ice.h
```diff
@@ -0,0 +1,38 @@
...
+ */
+#define PTA_EMMC_ICE_UUID \
+	{ 0x29e87b9e, 0x012a, 0x4878, \
+		{ 0xa1, 0xe1, 0xa1, 0xb9, 0x0a, 0x21, 0x5b, 0x16 } }
+
+/*
+ * Invalidate ICE key slot - overwrite key registers with random data
+ * [in]  params[0].value.a           Key slot index (0..ICE_MAX_KEY_IDX-1)
+ */
+#define PTA_CMD_ICE_INVALIDATE_KEY    2
```
harshaldev27: Why are we starting the PTA CMDs from 2? Why not start from 1? Are we trying to keep the same numbers as the SCM call?

core/pta/qcom/ice/sw_keys/ice_sw_keys.c
```diff
@@ -0,0 +1,281 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
harshaldev27, line 1: In my opinion, since we will soon add support for UFS ICE, it is better to place the PTA in a folder called 'ice', following by subfolder 'emmc'.

core/pta/qcom/ice/config.h
```diff
@@ -0,0 +1,59 @@
...
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_16_ADDR(r) \
+	(SDCC_ICE_LUT_KEYS + 0x40 + (0x80 * (r)))
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_16_CFGE_BMSK		0x80000000
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_16_CFGE_SHFT		31
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_16_CAPIDX_BMSK	0xff00
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_16_CAPIDX_SHFT	8
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_16_DUSIZE_BMSK	0xff
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_16_DUSIZE_SHFT	0
+
+#define ICE_MAX_KEY_IDX				32
```
harshaldev27, line 12: These configs seem agnostic to UFS/eMMC ICE controller, I suggest placing it in `core/pta/qcom/ice/config.h`. Only keep eMMC specific configs in this .h.

core/pta/qcom/emmc_ice/emmc_ice_config.c
```diff
@@ -0,0 +1,281 @@
...
+		EMSG("ICE invalidate: invalid slot %u (max %u)", index,
+		     ICE_MAX_KEY_IDX - 1);
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	ice_reg_write(SDC1_SDCC_ICE_CRYPTOCFG_r_16_ADDR(index), 0, 0, 0x0);
+
+	ice_wipe_key_regs(index);
+
+	ice_reg_write(SDC1_SDCC_ICE_CRYPTOCFG_r_16_ADDR(index),
```
harshaldev27: You should explain what we are doing in these three register writes.

core/pta/qcom/emmc_ice/emmc_ice_config.c
```diff
@@ -0,0 +1,281 @@
...
+	if (is_xts)
+		ice_write_key_to_regs(index, ICE_KEY_SALT_REG_START,
+				      key_buf + key_size, salt_size);
+	dsb();
+
+	return TEE_SUCCESS;
+}
+
+/* PTA command dispatcher */
+static TEE_Result invoke_command(void *sess_ctx __unused, uint32_t cmd_id,
```
harshaldev27: Can we create a file in `/core/pta/qcom/ice/` called `ice.c` and move these top-level entry points there? And keep only the emmc specific functions in this file, and then call them from the top-level file? This will help with extending ICE to UFS storage.

core/arch/arm/plat-qcom/main.c
```diff
@@ -28,6 +28,16 @@ register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICR_BASE,
 register_phys_mem_pgdir(MEM_AREA_IO_SEC, GICC_BASE, GIC_CPU_REG_SIZE);
 #endif
 
+#ifdef CFG_EMMC_ICE_FS_ENC_PTA
+/*
+ * Register the ICE (Inline Crypto Engine) LUT keys register region so the
+ * eMMC ICE PTA can program key slots.
+ */
+register_phys_mem_pgdir(MEM_AREA_IO_SEC,
```
harshaldev27: Hmm.. shouldn't we map this memory area statically in the PTA instead of in the platform init code? After all, OPTEE initialization doesn't need this memory, the ICE use-case requires this memory.

core/pta/qcom/ice/emmc/emmc_ice.h
```diff
@@ -0,0 +1,31 @@
...
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_16_ADDR(r) \
+	(SDCC_ICE_LUT_KEYS + 0x40 + (0x80 * (r)))
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_16_CFGE_BMSK		0x80000000
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_16_CFGE_SHFT		31
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_16_CAPIDX_BMSK	0xff00
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_16_CAPIDX_SHFT	8
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_16_DUSIZE_BMSK	0xff
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_16_DUSIZE_SHFT	0
+
+TEE_Result cmd_ice_invalidate_key(uint32_t param_types,
```
harshaldev27: Call them emmc_cmd_ice_*().

core/pta/qcom/ice/ice.c
```diff
@@ -0,0 +1,33 @@
...
+#include "emmc/emmc_ice.h"
+
+/* PTA command dispatcher */
+static TEE_Result invoke_command(void *sess_ctx __unused, uint32_t cmd_id,
+				 uint32_t param_types,
+				 TEE_Param params[TEE_NUM_PARAMS])
+{
+	switch (cmd_id) {
+	case PTA_CMD_ICE_INVALIDATE_KEY:
+		return cmd_ice_invalidate_key(param_types, params);
```
harshaldev27, line 40: Define these two in this file itself, and perhaps make them wrap emmc_cmd_ice*()? This way, we can add build time checks to call either emmc_cmd_ice*() or ufs_cmd_ice*() from these wrappers.

core/arch/arm/plat-qcom/bobcat/ipq52xx/target.mk
```diff
@@ -13,3 +13,6 @@ $(call force,CFG_WITH_SOFTWARE_PRNG,n)
 CFG_HWRNG_PTA ?= y
 CFG_HWRNG_QUALITY ?= 1024
 CFG_HWRNG_RATE ?= 0
+
+# Enable eMMC ICE filesystem-encryption PTA
+CFG_EMMC_ICE_FS_ENC_PTA ?= y
```
harshaldev27: This PTA is generic, it will be used for both UFS and eMMC, let's not call it CFG_EMMC_ICE_FS_ENC_PTA, call it CFG_ICE_FS_ENC_PTA.

lib/libutee/include/pta_qcom_ice.h
```diff
@@ -0,0 +1,38 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
```
harshaldev27, line 1: Same for this file, there's not need to call it an 'eMMC' specific PTA, this PTA will cater to all kind of ICE devices on different storage controllers.

harshaldev27: Also, this is a QCOM platform specific PTA, so let's call it `pta_qcom_ice.h`.

core/pta/qcom/ice/ice.c
```diff
@@ -0,0 +1,33 @@
...
+		break;
+	}
+
+	EMSG("ICE: Command not implemented: %u", cmd_id);
+	return TEE_ERROR_NOT_IMPLEMENTED;
+}
+
+pseudo_ta_register(.uuid = PTA_EMMC_ICE_UUID,
+		   .name = "emmc_ice.pta",
+		   .flags = PTA_DEFAULT_FLAGS | TA_FLAG_DEVICE_ENUM,
```
harshaldev27, line 66: `PTA_DEFAULT_FLAGS` makes it a multi session PTA, we will only ever have one driver talking to this PTA, multiple ICE driver instances talking to this PTA doesn't make sense, I think better just use `TA_FLAG_SINGLE_INSTANCE`.

thariaruchamy-bit: Kept `PTA_DEFAULT_FLAGS | TA_FLAG_DEVICE_ENUM` here, because `TA_FLAG_SINGLE_INSTANCE` on its own isn't a valid PTA flag set:

The framework *mandates* that full set. `verify_pseudo_tas_conformance()` in `core/kernel/pseudo_ta.c` calls `panic("PTA")` at init if `(flags & PTA_MANDATORY_FLAGS) != PTA_MANDATORY_FLAGS`. A bare `.flags = TA_FLAG_SINGLE_INSTANCE` (dropping `MULTI_SESSION`/`INSTANCE_KEEP_ALIVE`) would panic on every boot, so it isn't an option for a PTA. `MULTI_SESSION` here means multiple *sessions* against the one instance, not multiple instances,  every in-tree PTA carries it.

core/pta/qcom/ice/sw_keys/ice_sw_keys.c
```diff
@@ -0,0 +1,276 @@
...
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
harshaldev27, line 3: Hi @thariaruchamy-bit , It appears that we had side-stepped a bit. The functions and flow of ICE LUT keys programming when we are programming software generated keys is the same regardless of UFS/EMMC, what differs are the addresses/names of the registers in either case. I suggest we re-name this subfolder and files to `sw_keys/ice_sw_keys.c`, and also the prefixes to from `emmc` to `sw`. The UFS support can simply add the UFS controller specific register names to `ice_sw_keys.h` and any helper macros to decide between the two. The decision to invoke either functions in `ice_sw_keys.c` or in a latter `hwkm` folder will be runtime based, depending on the key size passed to the PTA. Do think about this, since it aligns with QTEE's approach.

core/pta/qcom/ice/ice.c
harshaldev27: Ack, let's keep it this way.

core/pta/qcom/ice/ice.c
```diff
@@ -0,0 +1,52 @@
...
+		return cmd_ice_set_config_key(param_types, params);
+	default:
+		break;
+	}
+
+	EMSG("ICE: Command not implemented: %u", cmd_id);
+	return TEE_ERROR_NOT_IMPLEMENTED;
+}
+
+pseudo_ta_register(.uuid = PTA_QCOM_ICE_UUID,
```
zelvam95, line 64: PTA has no `open_session_entry_point`; any normal-world client can program or wipe any ICE key slot
```c
pseudo_ta_register(.uuid = PTA_QCOM_ICE_UUID,
		   .name = "qcom_ice.pta",
		   .flags = PTA_DEFAULT_FLAGS | TA_FLAG_DEVICE_ENUM,
		   .invoke_command_entry_point = invoke_command);
```
No `open_session_entry_point` is provided. `pseudo_ta_enter_open_session()`
(`core/kernel/pseudo_ta.c:159-176`) only calls into a pseudo-TA's session
gate `if (stc->pseudo_ta->open_session_entry_point)` — when absent, session
open always succeeds, for any caller. `TA_FLAG_DEVICE_ENUM` only controls
whether the PTA is enumerable without tee-supplicant; it is not an
access-control gate.

The commit message explicitly states this PTA exists "to configure raw
software keys for hardware-accelerated storage encryption" driven by "the
kernel dm-crypt/inline-crypt path" — i.e. it is meant to be invoked from the
Linux kernel via the normal-world TEE client API, but it does not verify
the caller is that specific kernel driver, or reject a non-kernel/other
normal-world caller.

Contrast with the PTA this commit's own message says it "mirrors": the
existing `core/pta/qcom/pas/` PTA restricts every session to one specific
calling TA UUID:
```c
// core/pta/qcom/pas/pta_qcom_pas.c:177-193
pta_qcom_pas_open_session(uint32_t pt __unused, ...)
{
	struct ts_session *s = ts_get_calling_session();
	TEE_UUID ta_uuid = TA_PAS_UUID;
	...
	if (memcmp(&ctx->uuid, &ta_uuid, sizeof(TEE_UUID)))
		return TEE_ERROR_ACCESS_DENIED;
	return TEE_SUCCESS;
}
```
Without an equivalent gate here, **any process capable of opening a TEE
client session on the device** (whatever userspace/kernel component that
turns out to be, gated only by whatever OS-level permissions restrict
`/dev/tee0` — not by OP-TEE itself) can call `PTA_CMD_ICE_SET_CONFIG_KEY`
to inject attacker-controlled key material into any of the 32 ICE key
slots (`ICE_MAX_KEY_IDX`, `config.h:11`), or `PTA_CMD_ICE_INVALIDATE_KEY`
to wipe any slot — a disk-encryption integrity/availability compromise, not
merely a logic bug.

**Fix:** add an `open_session_entry_point` that restricts callers the same
way `pta_qcom_pas_open_session()` does (or via whatever mechanism is the
intended trust boundary for the real kernel ICE caller — e.g. checking
`ts_get_calling_session()`'s client identity/login type), matching the
precedent this PTA claims to mirror.

core/pta/qcom/ice/emmc/emmc_ice.c
```diff
@@ -0,0 +1,276 @@
...
+}
+
+/* Overwrite all data registers of key slot @index with random data. */
+static void ice_wipe_key_regs(uint32_t index)
+{
+	uint8_t rand_data[ICE_AES256_KEY_SIZE * 2];
+	int i;
+
+	if (crypto_rng_read(rand_data, sizeof(rand_data)) != TEE_SUCCESS)
+		memset(rand_data, 0xFF, sizeof(rand_data));
```
zelvam95: RNG failure falls back to a fixed, predictable wipe pattern
```c
if (crypto_rng_read(rand_data, sizeof(rand_data)) != TEE_SUCCESS)
	memset(rand_data, 0xFF, sizeof(rand_data));
```
If the RNG read fails, every "wiped" key register is set to the same
constant (`0xFF` repeated) rather than random data. This is a weaker
security posture than the function's stated intent ("Security wipe:
overwrite all 16 registers with random data") — a predictable wipe pattern
is a (minor) improvement over leaving stale key material, but is
distinguishable/predictable in a way true randomness is not. Consider
returning an error (and refusing to leave the slot enabled) instead of
silently downgrading to a fixed pattern, or at minimum log the fallback so
it's visible operationally.

How is it handled with QTEE? Are we mimicking QTEE here?

core/pta/qcom/ice/emmc/emmc_ice.c
```diff
@@ -0,0 +1,276 @@
...
+	 */
+	ice_configure_slot(index, ICE_CIPHER_MODE_XTS_256,
+			   ICE_DATA_UNIT_SIZE_512);
+	dsb();
+
+	return TEE_SUCCESS;
+}
+
+/*
+ * Command: tzbps_es_set_config_ice_key port — direct-copy path only.
```
zelvam95: comment "direct-copy path only" reads as if referencing a missing sibling path
```c
/*
 * Command: tzbps_es_set_config_ice_key port — direct-copy path only.
 * The kernel always passes raw plaintext key bytes on this platform.
 */
```
Searched the series for any "wrapped key"/"hardware-wrapped" counterpart
this comment might be contrasting with — none exists in this PR or
elsewhere in the tree. Read in isolation this could suggest a hardware-key-
blob path was intended but omitted; if that's not the case (i.e. this
platform genuinely only ever uses software keys), consider rewording to
state that plainly rather than "direct-copy path only," which implies an
alternative that doesn't exist.

Should we rather give a more generic description for ICE SW Key flow in Bobcat arch?

core/arch/arm/plat-qcom/bobcat/ipq52xx/target.mk
```diff
@@ -13,3 +13,6 @@ $(call force,CFG_WITH_SOFTWARE_PRNG,n)
 CFG_HWRNG_PTA ?= y
 CFG_HWRNG_QUALITY ?= 1024
 CFG_HWRNG_RATE ?= 0
+
+# Enable ICE filesystem-encryption PTA
+CFG_ICE_FS_ENC_PTA ?= y
```
harshaldev27, line 18: As discussed, you should define these macros in the `target_config.h` file and then use them in the ICE driver. You probably want to club the definition of these macros as part of the previous commit:

```
#define ICE_LUT_KEYS                 SDCC_ICE_LUT_KEYS
#define ICE_LUT_KEYS_SIZE            SDCC_ICE_LUT_KEYS_SIZE
```

core/pta/qcom/ice/emmc/emmc_ice.h
```diff
@@ -0,0 +1,31 @@
...
+#ifndef __EMMC_ICE_H
+#define __EMMC_ICE_H
+
+#include <platform_config.h>
+#include <tee_api_types.h>
+
+#include "../config.h"
+
+/* SDC1 SDCC ICE register definitions */
+#define SDC1_SDCC_ICE_CRYPTOCFG_r_n_ADDR(r, n) \
```
harshaldev27: As discussed, drop the prefix `SD1_SDCC` from these macros and move them to `config.h` so they can be used by the driver without worrying whether the controller being programmed is in UFS or eMMC.

core/pta/qcom/ice/emmc/emmc_ice.c
```diff
@@ -0,0 +1,276 @@
...
+
+/*
+ * Program CRYPTOCFG_r_16: cipher/capability mode (CAPIDX), the data-unit
+ * size for XTS tweak calculation (DUSIZE), and finally set CFGE to mark
+ * the slot's configuration valid/enabled.
+ */
+static void ice_configure_slot(uint32_t index, uint32_t cap_index,
+			       uint32_t data_unit_size)
+{
+	ice_reg_write(SDC1_SDCC_ICE_CRYPTOCFG_r_16_ADDR(index),
```
harshaldev27: Similarly, simply use `ICE_CRYPTOCFG..` everywhere.

core/pta/qcom/ice/ice.c
thariaruchamy-bit: Made the changes

core/pta/qcom/ice/emmc/emmc_ice.c
thariaruchamy-bit: Added log fallback for the failure instead of predictable pattern

core/pta/qcom/ice/emmc/emmc_ice.c
thariaruchamy-bit: Done

core/pta/qcom/ice/sw_keys/ice_sw_keys.c
```diff
@@ -0,0 +1,276 @@
...
+	/* Security wipe: overwrite all 16 registers with random data */
+	ice_wipe_key_regs(index);
+
+	/*
+	 * Re-arm the slot with a fixed, non-secret placeholder config
+	 * instead of leaving it disabled with stale configuration, so an
+	 * invalidated slot ends up in a deterministic, non-secret state.
+	 */
+	ice_configure_slot(index, ICE_CIPHER_MODE_XTS_256,
+			   ICE_DATA_UNIT_SIZE_512);
```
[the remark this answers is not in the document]
thariaruchamy-bit: Kept as-is by design: an invalidated slot is deliberately re-armed with a fixed, random-keyed config rather than left disabled. Leaving CFGE cleared would mean the slot has no active configuration at all, whereas re-arming with a placeholder cipher mode over freshly-wiped random key material keeps it in a well-defined, non-secret state that's safe to reuse. We did tighten the adjacent RNG-failure path: if the RNG read fails, the slot is now left disabled instead of being re-armed over an incomplete wipe (see the other thread on this file).

core/pta/qcom/ice/sw_keys/ice_sw_keys.c
```diff
@@ -0,0 +1,276 @@
...
+		key_size = ICE_AES256_KEY_SIZE; salt_size = ICE_AES256_KEY_SIZE;
+		is_xts = true;  break;
+	case ICE_CIPHER_MODE_CBC_256:
+	case ICE_CIPHER_MODE_ECB_256:
+		key_size = ICE_AES256_KEY_SIZE; salt_size = 0;
+		is_xts = false; break;
+	default:
+		EMSG("ICE set_key: unhandled cap_index %u", cap_index);
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
```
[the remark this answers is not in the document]
thariaruchamy-bit: Fixed: removed the two unreachable ICE_CIPHER_MODE_ECB_128/256 cases from this switch, since ECB is already rejected earlier in the function.

core/pta/qcom/ice/emmc/emmc_ice.c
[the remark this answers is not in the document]
thariaruchamy-bit: Not a typo: "tzbps_es_set_config_ice_key" is the exact name of the platform syscall this comment documents, which itself has that spelling ("tzbps" not "tzbsp"). Left as-is; "tzbsp_es_invalidate_ice_key" in the comment above it is spelled the standard way because that syscall's real name uses "tzbsp".

core/pta/qcom/sub.mk
```diff
@@ -1 +1,2 @@
 subdirs-$(CFG_QCOM_PAS_PTA) += pas
+subdirs-$(CFG_ICE_FS_ENC_PTA) += ice
```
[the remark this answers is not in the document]
thariaruchamy-bit: Fixed: the PR description was stale, referencing an earlier flag name from before a rename. Code has consistently used CFG_ICE_FS_ENC_PTA; updated the PR title/description to match.

core/pta/qcom/ice/sw_keys/ice_sw_keys.c
```diff
@@ -0,0 +1,292 @@
...
+
+	index = params[0].value.a;
+	if (index >= ICE_MAX_KEY_IDX) {
+		EMSG("ICE invalidate: invalid slot %u (max %u)", index,
+		     ICE_MAX_KEY_IDX - 1);
+		return TEE_ERROR_BAD_PARAMETERS;
+	}
+
+	/* Disable the slot so ICE can't use stale key data while it is wiped */
+	ice_reg_write(ICE_CRYPTOCFG_r_16_ADDR(index), 0, 0, 0x0);
```
harshaldev27: I notice the QTEE code adds a dsb() after this. Please check whether it is fine to omit the barrier.

thariaruchamy-bit: Checked and confirmed that per-write dsb() barriers are unnecessary here, since the ICE LUT-key region is mapped as Device-nGnRE and io_write32() already preserves the required write ordering.

The existing dsb() at the end of each command is sufficient to ensure all writes are completed before returning to the normal world, which also aligns with the established OP-TEE convention.

core/pta/qcom/ice/sw_keys/ice_sw_keys.c
```diff
@@ -0,0 +1,292 @@
...
+					   sizeof(uint32_t));
+
+	if (mask == 0 && shift == 0)
+		io_write32(va, val);
+	else
+		io_mask32(va, val << shift, mask);
+}
+
+/* Returns true if every byte of @key is zero. */
+static bool ice_key_is_all_zeros(const uint8_t *key, uint32_t key_len)
```
b49020, line 47: Instead of this helper, use `consttime_memcmp` for key based checks.

core/pta/qcom/ice/sw_keys/ice_sw_keys.c
```diff
@@ -0,0 +1,292 @@
...
+register_phys_mem_pgdir(MEM_AREA_IO_SEC,
+			(ICE_LUT_KEYS & ~SMALL_PAGE_MASK),
+			ICE_LUT_KEYS_SIZE);
+
+/* Write ICE register */
+static inline void ice_reg_write(paddr_t addr, uint32_t mask, uint32_t shift,
+				 uint32_t val)
+{
+	vaddr_t va = (vaddr_t)phys_to_virt(addr, MEM_AREA_IO_SEC,
+					   sizeof(uint32_t));
```
b49020: Don't do `phys_to_virt` on every register write. Rather just do once for the base address and then use it.

zelvam95:
[T0nyJH](https://github.com/T0nyJH) and [Jens Wiklander (@jenswikl)](https://github.com/jenswikl), Can you please share your thoughts on this one?^

thariaruchamy-bit:
@T0nyJH,  @jenswikl , Gentle remainder on this

b49020:
To be honest, I see especially in this inline encryption with software key proposal, from security aspect TZ has minimum or no role to play apart from providing an ABI to program the ICE IP with the key provided from non-secure world.

However, looks like as per access control policy only TZ can provision the ICE keys otherwise it would have been better to have a Linux ICE driver directly programming the ICE keys.

Anyhow, given the requirement to support fscrypt software key use-cases on Qcom SoCs (https://www.kernel.org/doc/html/latest/filesystems/fscrypt.html#filesystem-level-encryption-fscrypt), I won't mind exposing an OP-TEE pseudo TA interface given how the access control policy is configured.

@ldts let me know if you still have any further objections to this interface.

b49020:
Please rebase this PR to tip of `qcom-next`.

thariaruchamy-bit:
> Please rebase this PR to tip of `qcom-next`.

Rebase done @b49020

ldts:
@b49020 I would still have preferred to inject the key during board provisioning. Seems the logical thing to promote.

Regardless, we could address the same TOCTOU issue that was reported in TF-A a few months ago via CVE if you remember (Elias reported it). and so:

1) use an intermediate secure buffer in the stack in sw_cmd_ice_set_config_key; so should memcopy params[2].memref.buffer to it and continue working from the stack instead. 

2) then just memzero_explicit the intermediate buffer on exit.

with that 

`Reviewed-by: Jorge Ramirez-Ortiz <jorge.ramirez@oss.qualcomm.com>`

zelvam95:
Based on above review comments + prior reviews, you can add my tag as well.

`Reviewed-by: Selvam Sathappan Periakaruppan <speriaka@qti.qualcomm.com>
`

harshaldev27:
`Reviewed-by: Harshal Dev <harshal.dev@oss.qualcomm.com>`

thariaruchamy-bit:
@ldts Comments are resolved, added tags.
@b49020 Can you please help merge this PR into qcom-next?

FYI : Raised changes in upstream master branch as well (https://github.com/OP-TEE/optee_os/pull/7906) currently in draft, will request for review once the kernel patches are posted in upstream

PR #32, opened 2026-07-29
plat-qcom: ipq96xx: add CDSP (Turing/NSP) PAS remoteproc support

thariaruchamy-bit:
## Summary

Add PAS support for the ipq96xx CDSP (Turing/NSP) remote processor and
enable it on the ipq96xx target. Slots CDSP into the existing per-platform
PAS descriptor/ops framework already used by kodiak and lemans.

## Testing

- Built `PLATFORM_FLAVOR=ipq96xx` (feature on) and `=kodiak` (feature off)
  with `aarch64-linux-gnu-`; each commit builds independently
  (bisect-verified).
- checkpatch clean (0/0) on both commits.
- CDSP boot verified on ipq96xx hardware.

## Known limitation

No secure image authentication, matching the rest of the upstream PAS
framework; integrity relies on the loader chain.


core/pta/qcom/pas/platform/ipq96xx/cdsp.c
```diff
@@ -0,0 +1,293 @@
...
+		return TEE_ERROR_TIMEOUT;
+	}
+
+	DMSG("CDSP boot completed successfully");
+	dsb();
+
+	return TEE_SUCCESS;
+}
+
+static TEE_Result cdsp_fw_shutdown(struct qcom_pas_data *data)
```
zelvam95, line 188: `cdsp_fw_shutdown()` — two of three timeout waits proceed unchecked, unlike the boot path in the same file
```c
/* Wait for power-on status. */
timeout = timeout_init_us(1000 * 1000);
do {
	val = io_read32(tcsr + TCSR_TURING_PWR_ON);
	if (val & 0x1)
		break;
} while (!timeout_elapsed(timeout));
/* <- val's outcome is never checked; falls through to HALTREQ regardless */

/* Request halt and wait for acknowledgment. */
io_write32(tcsr + TCSR_TURING_HALTREQ, 0x1);
timeout = timeout_init_us(5000 * 1000);
do { ... } while (!timeout_elapsed(timeout));
if (!(val & 0x1)) {                              /* <- this one IS checked */
	EMSG("Failed to receive halt acknowledgment");
	return TEE_ERROR_TIMEOUT;
}
...
/* Wait for idle. */
timeout = timeout_init_us(5000 * 1000);
do {
	val = io_read32(tcsr + TCSR_TURING_MASTER_IDLE);
	val2 = io_read32(tcsr + TCSR_TURING_IL1_MASTER_IDLE);
	if ((val & 0x1) && (val2 & 0x1))
		break;
} while (!timeout_elapsed(timeout));
/* <- also never checked; function returns TEE_SUCCESS regardless */

return TEE_SUCCESS;
```
The first wait ("power-on status") and the last wait ("idle") never check
whether the loop actually observed the expected bit or just timed out —
execution proceeds into the halt-request/reset sequence and returns
`TEE_SUCCESS` either way. Only the middle wait (`HALTACK`) is checked and
returns `TEE_ERROR_TIMEOUT` on failure. This is an inconsistency within the
same function, not a project-wide pattern: `cdsp_fw_start()` in the same
file (lines 157-186) checks *both* of its boot-stage timeouts and returns
`TEE_ERROR_TIMEOUT` with an `EMSG` on failure for each.

Practical effect: if the subsystem was never actually powered on, the code
still issues a halt request and proceeds through the full GCC reset
sequence against a peripheral that may not be in the state the sequence
assumes. If the final idle wait times out, the function still reports
`TEE_SUCCESS` to its caller even though the subsystem may not have reached
a safe idle state before the AHBS clock is toggled and reset lines are
deasserted afterward — a caller (or a subsequent re-bringup attempt) has no
way to know the shutdown didn't fully complete as expected.

**Suggest:** check both loops the same way the boot path does — log and
return `TEE_ERROR_TIMEOUT` (or at least log a warning) if power-on status
or final idle status was never observed, matching `cdsp_fw_start()`'s
pattern in the same file.

core/pta/qcom/pas/platform/pas_data.h
```diff
@@ -16,6 +16,7 @@
 #define PAS_ID_IRIS		9
 #define PAS_ID_TURING		18
 #define PAS_ID_TURING1		30
+#define PAS_ID_CDSP_DTB		37
```
zelvam95: Should we update as TURING_DTB? Given the ROOTPD is referred to as PAS_ID_TURING?

core/pta/qcom/pas/platform/ipq96xx/cdsp.c
```diff
@@ -0,0 +1,293 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+/*
+ * ipq96xx CDSP (Turing/NSP) bring-up.
+ *
+ * NOTE: the CDSP firmware and its device-tree blob are booted here without
```
zelvam95: I think we can completely remove this comment? We already have a to-do else where for all targets to plug-in auth I think?

core/drivers/clk/qcom/platform/ipq96xx/clock-qcom-pas.c
```diff
@@ -0,0 +1,106 @@
...
+		   CDSPAUX_BRIDGE_DELAY_CYCLES);
+}
+
+TEE_Result qcom_clock_enable_pas_processor(enum qcom_clk_group group __unused)
+{
+	/* The DSP core is released entirely within the PTA fw_start path. */
+	return TEE_SUCCESS;
+}
+
+TEE_Result qcom_clock_pas_reset(enum qcom_clk_group group __unused)
```
zelvam95, line 72: Is this how its handled for other co-processors? Cross check and see if there is any scope for alignment for both this function and the enable_pas_processor function? Currently both of those seem to just return tee_success?

core/pta/qcom/pas/platform/ipq96xx/cdsp.c
thariaruchamy-bit: Fixed in the latest push (commit `pta: qcom: pas: ipq96xx: add CDSP remoteproc bring-up`). `cdsp_fw_shutdown()` now:
- skips the halt request entirely when the power-on wait times out (subsystem already off), instead of issuing HALTREQ against a peripheral that was never powered on
- checks the final idle wait and returns `TEE_ERROR_TIMEOUT` with an `EMSG` if the subsystem never reached idle, matching `cdsp_fw_start()`'s pattern in the same file

The physical reset steps still run either way (can't safely abort mid-shutdown), but the function now reports failure accurately instead of always returning `TEE_SUCCESS`.

core/drivers/clk/qcom/platform/ipq96xx/clock_group_qcom.h
```diff
@@ -0,0 +1,77 @@
...
+#define _CLOCK_GROUP_QCOM_H_
+
+#include <util.h>
+
+/*
+ * Turing sub-block offsets within the TURING controller window
+ * (TURING_BASE/TURING_SIZE). The window spans the Turing clock controller and
+ * the QDSP6 subsystem, mapped once by the PAS PTA (pas_platform_mem_setup); the
+ * clock driver derives sub-block bases as TURING_BASE + offset.
+ */
```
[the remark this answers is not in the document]
thariaruchamy-bit: Fixed, corrected the comment to describe the driver's own mapping instead of the PTA's.

core/pta/qcom/pas/platform/ipq96xx/cdsp.c
```diff
@@ -0,0 +1,293 @@
...
+	vaddr_t qdsp6ss = 0;
+	vaddr_t mpm2_va = 0;
+	vaddr_t spare_va = 0;
+	uint32_t boot_status = 0;
+	uint32_t boot_vector_evb = 0;
+	uint64_t timeout = 0;
+	int debug_q6 = 0;
+
+	if (!turing || !dtb)
+		return TEE_ERROR_BAD_STATE;
```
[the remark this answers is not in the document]
thariaruchamy-bit: Fixed in the latest push: `cdsp_fw_start()` now rejects with `TEE_ERROR_NO_DATA` if `dtb->fw_base`/`dtb->fw_size` aren't set, before programming the DTB config registers.

core/pta/qcom/pas/platform/ipq96xx/cdsp.c
```diff
@@ -0,0 +1,293 @@
...
+		if (val & 0x1)
+			break;
+	} while (!timeout_elapsed(timeout));
+
+	if (!(val & 0x1)) {
+		EMSG("Failed to receive halt acknowledgment");
+		return TEE_ERROR_TIMEOUT;
+	}
+
+	io_write32(turing_cc + TURING_CC_ALT_RESET_CTL, 0x1);
```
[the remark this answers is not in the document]
thariaruchamy-bit: Fixed with the same change requested by @zelvam95 above: the power-on wait now skips the halt request when the subsystem is already off, and the idle wait is now checked and returns `TEE_ERROR_TIMEOUT` on failure, matching `cdsp_fw_start()`'s pattern.

core/drivers/clk/qcom/platform/ipq96xx/clock-qcom-pas.c
thariaruchamy-bit: Checked: this already matches the existing convention. kodiak's `clock-qcom-pas.c` has the identical pattern, both `qcom_clock_enable_pas_processor()` and `qcom_clock_pas_reset()` are `__unused`-parameter no-ops returning `TEE_SUCCESS`, for the same reason (the DSP core is released entirely within the PTA `fw_start` path, so there's nothing left for these hooks to do).

Also worth noting: since ipq96xx's TURING subsystem uses `reset_seq = QCOM_PAS_RESET_CLK_ENABLE` (same as kodiak), `pas_core.c`'s `pas_platform_auth_and_reset()` never actually calls either of these two functions for that reset_seq (only `QCOM_PAS_RESET_CLK_FULL`, used by lemans, calls them). So on ipq96xx they're currently unreachable/dead code, kept only to satisfy the `clk_qcom.h` interface, same as on kodiak. Leaving them as-is for consistency with kodiak; let me know if you'd rather we drop them from both platforms instead.

core/pta/qcom/pas/platform/ipq96xx/cdsp.c
```diff
@@ -0,0 +1,293 @@
...
+	io_write32(qdsp6ss + Q6SS_BOOT_CORE_START_REG, BOOT_CORE_START_ENABLE);
+	io_write32(qdsp6ss + Q6SS_BOOT_AUTO_BREAK_EN_REG,
+		   BOOT_AUTO_BREAK_ENABLE);
+
+	/* Configure the device-tree blob. */
+	io_write32(qdsp6ss + DTB_CONFIG_0_REG, (uint32_t)dtb->fw_base);
+	io_write32(qdsp6ss + DTB_CONFIG_1_REG, (uint32_t)(dtb->fw_base >> 32));
+	io_write32(qdsp6ss + DTB_CONFIG_2_REG, DTB_CHIP_FAMILY_ID);
+	io_write32(qdsp6ss + DTB_CONFIG_3_REG, DTB_VERSION);
+	io_write32(qdsp6ss + DTB_CONFIG_5_REG, dtb->fw_size);
```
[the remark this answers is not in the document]
thariaruchamy-bit: Fixed, added the explicit `(uint32_t)` cast.

core/pta/qcom/pas/platform/ipq96xx/cdsp.c
thariaruchamy-bit: done

core/pta/qcom/pas/platform/pas_data.h
thariaruchamy-bit: done

core/pta/qcom/pas/platform/ipq96xx/subsys.c
```diff
@@ -0,0 +1,40 @@
...
+#include <stddef.h>
+#include <util.h>
+
+#include "cdsp.h"
+#include "pas_subsys.h"
+
+static struct qcom_pas_subsys subsystems[] = {
+	{
+		.data = {
+			.pas_id = PAS_ID_TURING_DTB,
```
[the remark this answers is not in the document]
ldts: > No, there's no such slot count in this implementation. The subsystem list is a plain static array returned via `ARRAY_SIZE(subsystems)`, so adding a `PAS_ID` just adds an entry.

that is obviously not what I meant...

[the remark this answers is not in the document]
ldts: > Jorge A. Ramirez-Ortiz (@ldts), For Bobcat, TMEL based auth is used and PAS_AUTH config will not be enabled in these targets. A new config where TMEL Based Auth is impl would be added/enabled once TMEL-COM support is added. (Its WIP currently)

ok. that is what I was after. thanks @zelvam95

core/pta/qcom/pas/platform/ipq96xx/cdsp.c
```diff
@@ -0,0 +1,306 @@
...
+		return TEE_ERROR_TIMEOUT;
+	}
+
+	DMSG("CDSP boot completed successfully");
+	dsb();
+
+	return TEE_SUCCESS;
+}
+
+static bool cdsp_powered_on(vaddr_t tcsr)
```
ldts: better do a couple of functions `bool cdsp_state_sync (STATE)` and `cdsp_state_enter(STATE)` or something like that, makes it more readable

core/pta/qcom/pas/platform/ipq96xx/subsys.c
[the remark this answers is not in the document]
ldts: can we **`force`** authentication to N on the architecture so it is explicit please

core/pta/qcom/pas/platform/ipq96xx/cdsp.c
thariaruchamy-bit: Done, made the changes

core/pta/qcom/pas/platform/ipq96xx/subsys.c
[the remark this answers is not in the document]
thariaruchamy-bit: Sure, have made the changes

thariaruchamy-bit:
@ldts Have addressed the above comments, please let us know if there are any additional ones

thariaruchamy-bit:
@ldts gentle remainder on this

b49020:

thariaruchamy-bit:
> Please rebase this PR to tip of qcom-next.

Rebase done @b49020

b49020:
Upstream PR link? Also, fix the merge conflict.

core/pta/qcom/pas/platform/ipq96xx/cdsp.c
```diff
@@ -0,0 +1,324 @@
...
+
+	dsb();
+	isb();
+
+	timeout = timeout_init_us(BOOT_TIMEOUT_MS * 1000);
+	do {
+		boot_status = io_read32(qdsp6ss + Q6SS_BOOT_STATUS_REG);
+		if (boot_status == Q6SS_BOOT_STATUS_STAGE1)
+			break;
+	} while (!timeout_elapsed(timeout));
```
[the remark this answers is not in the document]
zelvam95: @thariaruchamy-bit, We could check and fix this? Seems to be valid; In couple of places we need to add this delay looks like as per above comment; Please check and add it;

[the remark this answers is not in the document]
thariaruchamy-bit: Addressed the comments

b49020:
@thariaruchamy-bit please do a clean rebase on top of `qcom-next` dropping the merge commit.

thariaruchamy-bit:
> Harikrishna (@thariaruchamy-bit) please do a clean rebase on top of `qcom-next` dropping the merge commit.

Sure Sumit, have rebased cleanly now

@b49020 @ldts @zelvam95 Upstream PR created : https://github.com/OP-TEE/optee_os/pull/8000

core/arch/arm/plat-qcom/bobcat/ipq96xx/target.mk
```diff
@@ -10,5 +10,17 @@ CFG_TZDRAM_START ?= 0x8A680000
...
 CFG_TEE_RAM_VA_SIZE ?= 0x280000
 
+CFG_DRIVERS_CLK ?= y
+CFG_DRIVERS_QCOM_CLK ?= y
+
+CFG_QCOM_PAS_PTA ?= y
+
+ifeq ($(CFG_QCOM_PAS_PTA),y)
+# The CDSP controller window is mapped at runtime from the reserved VA pool.
+CFG_RESERVED_VASPACE_SIZE ?= (16 * 1024 * 1024)
```
b49020: Is this really needed now since we have the dynamic VA SPACE allocation merged upstream?

core/drivers/clk/qcom/platform/ipq96xx/clock-qcom-pas.c
b49020: Please check the rename happened for these files for other platforms. Lets be consistent.

core/drivers/clk/qcom/platform/ipq96xx/clock_group_qcom.h
b49020: Ditto here.

core/arch/arm/plat-qcom/bobcat/ipq96xx/target.mk
thariaruchamy-bit: removed the change

thariaruchamy-bit:
@b49020 Have resolved the comments, please let me know if there are any additional ones

zelvam95:
> Sumit Garg (@b49020) Have resolved the comments, please let me know if there are any additional ones

Hi Hari

As we synced, let’s test once with the latest changes/PR to ensure remoteproc start/stop can work by trying it for couple of times.

thariaruchamy-bit:
> > Sumit Garg (Sumit Garg (@b49020)) Have resolved the comments, please let me know if there are any additional ones
> 
> Hi Hari
> 
> As we synced, let’s test once with the latest changes/PR to ensure remoteproc start/stop can work by trying it for couple of times.

Tried the remote proc bring up sequence multiple times, the bring up was successful

b49020:
Overall looks good to merge, the rest of the review can happen upstream.

PR #33, opened 2026-07-31
drivers: qcom: add GENI SPI driver and enable spi config

VeshalaAnilKumar:
Add a Qualcomm SPI geni driver implementing spi_ops for a GENI Serial Engine in FIFO transfer mode with interrupt-driven. Enabled CFG_QCOM_GENI_SPI in lemans platform, and add the corresponding qup config settings in qup_spi_config[] table.


core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,868 @@
...
+
+	/* Transfer complete: gate the SE source clock again. */
+	qup_spi_clk_disable(qs);
+}
+
+static void qup_spi_flushfifo(struct spi_chip *chip)
+{
+	struct qup_spi_data *qs = to_qup_spi(chip);
+
+	while (io_read32(qs->base + SE_GENI_RX_FIFO_STATUS) & RX_FIFO_WC_MSK)
```
zelvam95, line 549: unbounded busy-waits on RX FIFO and M_GENI_CMD_ACTIVE; a stuck FIFO or hung SE hangs this call forever

```c
while (io_read32(qs->base + SE_GENI_RX_FIFO_STATUS) & RX_FIFO_WC_MSK)
    io_read32(qs->base + SE_GENI_RX_FIFOn);

while (io_read32(qs->base + SE_GENI_STATUS) &
       GENI_STATUS_M_GENI_CMD_ACTIVE)
    ;
```

Both loops have no timeout. If the SE is in an error state (SW_ERR
that the driver doesn't yet check, bus fault, clock disabled), this
hangs OP-TEE indefinitely with interrupts disabled by the caller (any
`spi_ops.flushfifo` caller from SPI framework code). A stuck-forever
loop in secure world is significantly worse than a stuck-forever loop
in normal world.

The rest of the driver already uses the `timeout_init_us(...)` +
`timeout_elapsed(...)` idiom (see `qup_spi_wait_flag`); apply the same
bound here.

**Suggest:** cap both loops with `timeout_init_us(QUP_SPI_M_CMD_TIMEOUT_US)`
and `EMSG(...)` + return early on timeout. A caller can then observe
the failure via the next `txrx` attempting to error out cleanly.

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,868 @@
...
+	*dfs_idx = best->dfs_idx;
+	*clk_div = best_div;
+
+	DMSG("QUP SPI: speed=%u Hz -> src=%u Hz div=%u dfs_idx=%u sclk=%u Hz",
+	     speed_hz, best->freq_hz, best_div, best->dfs_idx, best_sclk);
+
+	return TEE_SUCCESS;
+}
+
+static void qup_spi_configure(struct spi_chip *chip)
```
zelvam95, line 318: `qup_spi_configure()` — returns `void`; on `qcom_clk_get_domain()` / `geni_spi_resolve_clk()` failure the SE runs with **stale clock configuration** and the next transfer produces wrong SPI SCLK

```c
domain = qcom_clk_get_domain(qs->se_clk);
if (!domain) {
    EMSG("QUP SPI: no clock domain for SE %u", qs->id);
    return;              /* <-- SE not reconfigured */
}
res = geni_spi_resolve_clk(domain, qs->speed_hz, &dfs_idx, &clk_div);
if (res) {
    EMSG("QUP SPI: no DFS source rate for %u Hz: %#" PRIx32, ...);
    return;              /* <-- SE not reconfigured */
}
```

`configure()`'s `spi_chip::ops` signature returns `void`, so the driver
has no way to propagate the failure back to the caller. But this means
the caller's next `.start()` / `.txrx()` will run against whatever
clock state the SE was in before configure was called — potentially
the wrong CS, mode, packing, or bus speed. For a security-relevant SPI
target (e.g. an SPI-attached secure element or fTPM), running at a
speed the target rejects is a subtle failure mode; running at a speed
the target *does* accept but at the wrong CPHA/CPOL corrupts every
byte on the wire.

The safer contract is to set a `qs->configured = true` flag only on
full success, and to reject `start`/`txrx` when it is false.

**Suggest:** track configuration success in `struct qup_spi_data`
(e.g. `bool configured`), set it at the end of a fully-successful
`qup_spi_configure()`, clear it on any early-return, and short-circuit
`qup_spi_start` / `qup_spi_txrx` with an EMSG + no-op when the flag is
false.

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,868 @@
...
+
+	while (io_read32(qs->base + SE_GENI_RX_FIFO_STATUS) & RX_FIFO_WC_MSK)
+		io_read32(qs->base + SE_GENI_RX_FIFOn);
+
+	while (io_read32(qs->base + SE_GENI_STATUS) &
+	       GENI_STATUS_M_GENI_CMD_ACTIVE)
+		;
+}
+
+static enum spi_result qup_spi_txrx(struct qup_spi_data *qs,
```
zelvam95, line 557: AI Agents review comment on this function is as below. Can you please review this feedback and if applicable/plan to handle it?
## 🟡 `core/drivers/spi/qcom/qcom_geni_spi.c:557-605` `qup_spi_txrx()` — for RX-only transfers (`wdat == NULL`), `SE_SPI_TX_TRANS_LEN` is still programmed to `num_pkts` but the TX FIFO is never filled and the TX watermark is never armed

```c
io_write32(qs->base + SE_SPI_TX_TRANS_LEN, num_pkts & TRANS_LEN_MSK);   /* :568 */
io_write32(qs->base + SE_SPI_RX_TRANS_LEN, num_pkts & TRANS_LEN_MSK);   /* :569 */
...
qs->tx_rem_bytes = wdat ? total_bytes : 0;                              /* :574 */
...
if (qs->tx_rem_bytes) {                                                 /* :580 */
    io_write32(qs->base + SE_GENI_TX_WATERMARK_REG, 1);
    qup_spi_fifo_fill_tx(qs);
}
```

SPI is full-duplex — every clock cycle shifts a bit out on MOSI. For
an RX-only transfer, the SE must still be given *something* to
shift out (usually zeros). This driver programs `SE_SPI_TX_TRANS_LEN
= num_pkts` (asserting to the SE that `num_pkts` TX words will be
provided) but then never fills the TX FIFO, so the SE will either
stall on `TX_FIFO_WATERMARK` (never rearmed) or drive whatever
uninitialised state the FIFO happened to contain onto MOSI. Either
outcome is a correctness bug for pure-RX transfers.

Two typical fixes: (a) for RX-only, set `SE_SPI_TX_TRANS_LEN = 0` and
mask the TX-watermark IRQ; or (b) for RX-only, treat the transfer as
"transmit zeros" — fill the TX FIFO with 0-filled words until
`num_pkts` are shifted out.

**Suggest:** confirm the SE behaviour when `TX_TRANS_LEN > 0` and no
TX FIFO fill is provided (this is documented in the QUP SE HAS), then
fix `qup_spi_txrx` to program `TX_TRANS_LEN` only when `wdat != NULL`,
or fill the TX FIFO with zeros. Similarly, TX-only transfers should
program `RX_TRANS_LEN = 0` and disable the RX-watermark IRQ.

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,868 @@
...
+	struct qup_spi_data *qs = h->data;
+	uint32_t exceptions = cpu_spin_lock_xsave(&qs->lock);
+
+	qup_spi_isr_handle_fifo(qs);
+	cpu_spin_unlock_xrestore(&qs->lock, exceptions);
+
+	return ITRR_HANDLED;
+}
+
+static void qup_spi_start(struct spi_chip *chip)
```
zelvam95, line 499: `qup_spi_start()`/`qup_spi_end()` — clock ungate errors in start are dropped silently; end still deasserts CS and gates the clock

```c
static void qup_spi_start(struct spi_chip *chip)
{
    ...
    if (qup_spi_clk_enable(qs))
        return;    /* <-- start() returns void, error is lost */
    ...
    geni_setup_m_cmd(qs, SPI_CS_ASSERT, 0);
    ...
}
```

If `qup_spi_clk_enable()` fails, `qup_spi_start` returns early — CS
is never asserted, the caller's subsequent `.txrx*` runs against a
gated SE (the M_CMD write will land against a clock-gated peripheral),
and `.end()` will try to deassert CS that was never asserted. Same
observation as `configure()` (item 5): the `spi_ops` op is `void`, so
there is no error propagation. A `qs->started` flag gated on full
success of `start()` would short-circuit `txrx`/`end` when start
failed.

`spi_ops.start`/`.end`/`.configure` all being void is an OP-TEE
framework limitation, but this driver can defend against it with a
single boolean.

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,868 @@
...
+		DMSG("QUP SPI %u: base=%#" PRIxVA " irq=%zu se_clk=%s",
+		     qup_spi_id, qs->base, qs->itr_num, qs->se_clock_name);
+		return true;
+	}
+
+	EMSG("QUP SPI: no platform data for SE id %u", qup_spi_id);
+	return false;
+}
+
+TEE_Result qup_spi_init(struct qup_spi_data *qs, unsigned int qup_spi_id)
```
zelvam95, line 822: `qup_spi_init()` — `qs->id` used by `qup_spi_get_platform_data()`'s trace before being assigned

```c
if (!qup_spi_get_platform_data(qup_spi_id, qs))     /* uses qs->id internally? */
    return TEE_ERROR_ITEM_NOT_FOUND;

qs->id = qup_spi_id;                                /* <-- assigned here */
```

`qup_spi_get_platform_data()` at line 813 does:
```c
DMSG("QUP SPI %u: base=%#" PRIxVA " irq=%zu se_clk=%s",
     qup_spi_id, qs->base, qs->itr_num, qs->se_clock_name);
```
Note it uses the parameter `qup_spi_id`, not `qs->id` — so this is
actually fine. But `qup_spi_clk_setup()` (called after
`qup_spi_get_platform_data` but before `qs->id = qup_spi_id`) does
`EMSG("QUP SPI: no se clock name for SE %u", qs->id)` — printing 0
instead of the actual SE id. Consequential only for log messages, but
easy to fix by assigning `qs->id = qup_spi_id` **before**
`qup_spi_get_platform_data()`, or by having `get_platform_data` do the
assignment itself.

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,868 @@
...
+ */
+static void geni_config_packing(struct qup_spi_data *qs, unsigned int bpw)
+{
+	uint32_t cfg[NUM_PACKING_VECTORS] = { 0 };
+	int idx_start = bpw - 1;
+	int idx = idx_start;
+	int idx_delta = -8;
+	int temp_bpw = bpw;
+	int ceil_bpw = ROUNDUP(bpw, 8);
+	int iter = (ceil_bpw * 1) / 8;
```
zelvam95, line 211: `geni_config_packing()` — silent no-op if `bpw < 1` or `bpw > 32`; also `int iter = (ceil_bpw * 1) / 8;` has a stray `* 1`

```c
int ceil_bpw = ROUNDUP(bpw, 8);
int iter = (ceil_bpw * 1) / 8;
```
Two nits:
- `(ceil_bpw * 1) / 8` — the `* 1` reads like a leftover from a
  templated formula (maybe scaled per byte-vector count?). It reduces
  to `ceil_bpw / 8`. Either drop `* 1` or add a comment explaining why
  it's there.
- The guard `if (iter <= 0 || iter > NUM_PACKING_VECTORS) return;`
  silently no-ops for invalid `bpw`. Since `configure()` already
  `assert()`s `bpw >= MIN_WORD_LEN && bpw <= 32`, this can only be
  reached in a release build with the assert compiled out — in that
  case a silent no-op is worse than an EMSG.

---

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,868 @@
...
+	io_write32(qs->base + SE_GENI_TX_PACKING_CFG0,
+		   cfg[0] | (cfg[1] << PACKING_VECTOR_SHIFT));
+	io_write32(qs->base + SE_GENI_TX_PACKING_CFG1,
+		   cfg[2] | (cfg[3] << PACKING_VECTOR_SHIFT));
+	io_write32(qs->base + SE_GENI_RX_PACKING_CFG0,
+		   cfg[0] | (cfg[1] << PACKING_VECTOR_SHIFT));
+	io_write32(qs->base + SE_GENI_RX_PACKING_CFG1,
+		   cfg[2] | (cfg[3] << PACKING_VECTOR_SHIFT));
+
+	if (bpw == 32)
```
zelvam95, line 243: `geni_config_packing()` — `SE_GENI_BYTE_GRAN` written **only** when `bpw == 32`

```c
if (bpw == 32)
    io_write32(qs->base + SE_GENI_BYTE_GRAN, bpw / 16);
```
The Linux kernel GENI SE driver programs `BYTE_GRAN` for **any**
`bpw > 8` (see `drivers/spi/spi-geni-qcom.c` in mainline). Confirm
against the QUP SE hardware spec that `bpw > 16` on this SoC does not
require `BYTE_GRAN != default`; if it does, this branch is missing
cases 17..31.

core/include/drivers/qcom_geni_spi.h
```diff
@@ -0,0 +1,129 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2026, Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#pragma once
```
zelvam95, line 6: header uses both `#pragma once` **and** `#ifndef` guards; and the guard is `__DRIVERS_QCOM_GENI_SPI_H` (leading-underscore identifier reserved by C standard to the implementation)

```c
#pragma once

#ifndef __DRIVERS_QCOM_GENI_SPI_H
#define __DRIVERS_QCOM_GENI_SPI_H
```

OP-TEE headers uniformly use plain include guards (e.g.
`DRIVERS_QCOM_GENI_SPI_H`) without `#pragma once`. Two of the three
lines here are redundant, and the guard uses a name reserved to the
implementation (`__` prefix). Either delete `#pragma once` and rename
the guard to `DRIVERS_QCOM_GENI_SPI_H`, or drop the `#ifndef` pair
and keep only `#pragma once` — but the repo convention is the former.

---

core/drivers/spi/qcom/sub.mk
```diff
@@ -0,0 +1,8 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
zelvam95, line 1: — SPDX header is `BSD-3-Clause` while every other file in the PR is `BSD-2-Clause`

```make
# SPDX-License-Identifier: BSD-3-Clause
```
All other new files in this PR (`qcom_geni_spi.c`, `qcom_geni_spi.h`,
`qcom_geni_spi_config.c`, `platform/lemans/sub.mk`, `platform/sub.mk`)
are BSD-2-Clause. Sub-make files inherit the surrounding directory's
licence — this appears to be a copy-paste artefact from a different
subsystem. Change to `BSD-2-Clause` for consistency.

Also, check if sub.mk really requires copyright markings? Check other existing sub.mk files for ref.

---

core/drivers/spi/qcom/platform/lemans/qcom_geni_spi_config.c
```diff
@@ -0,0 +1,90 @@
...
+	"gcc_qupv3_wrap_2_m_ahb_clk",
+	NULL,
+};
+
+/*
+ * QUP SPI2 pads:
+ *   gpio86 = MISO, gpio87 = MOSI, gpio89 = CS -> pull-down, 6 mA
+ *   gpio88 = CLK                              -> no pull,   6 mA
+ * QUP_SPI2_PIN_FUNC is this SoC's TLMM function-mux value that selects
+ * the SPI SE on these pads -- set it to the real value for your target.
```
zelvam95, line 35: -- set it to the real value for your target.

This part of it should we remove from this file? Its in a target specific file, someone else enabling this needs to ensure to define this macro for the respective targets? That can be taken as ref.? Explicit comment needed?

b49020:
Please rebase to tip of `qcom-next`

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,1494 @@
...
+	struct qup_spi_data *qs = to_qup_spi(chip);
+	uint32_t exceptions = 0;
+
+	if (!qs->started) {
+		EMSG("QUP SPI %u: end() called without a successful start()",
+		     qs->id);
+		return;
+	}
+
+	exceptions = cpu_spin_lock_xsave(&qs->lock);
```
b49020: Please avoid expensive CPU spinlocks, instead switch to using mutex throughout this driver.

b49020:
@VeshalaAnilKumar were you able to test dTPM communication over SPI on Lemans EVK using this driver?

VeshalaAnilKumar:
> VeshalaAnilKumar were you able to test dTPM communication over SPI on Lemans EVK using this driver?

tested only spi loop back , dTPM use case clients can verify it

core/drivers/spi/qcom/platform/lemans/qcom_geni_spi_config.c
```diff
@@ -0,0 +1,1383 @@
...
+#include <mm/core_mmu.h>
+#include <util.h>
+
+/*
+ * Per-QUPv3 wrapper resources. The wrapper-common register window and the
+ * wrapper gating clocks are shared by every SE in that wrapper, so they are
+ * defined and mapped once here rather than per SE.
+ */
+
+const uint8_t spi_qup_fw[] =
```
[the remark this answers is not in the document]
b49020: Please address this comment.

core/drivers/spi/qcom/qcom_geni_spi_test.c
```diff
@@ -0,0 +1,180 @@
...
+		     qup_spi_config[0].id, res);
+		return TEE_SUCCESS;
+	}
+
+	if (!qup_spi_test_loopback(&qs))
+		EMSG("QUP SPI %u: loopback test: FAIL", qs.id);
+
+#ifdef CFG_QUP_SPI_TEST_EXTERNAL
+	qup_spi_test_external(&qs);
+#endif
```
b49020: I don't think we should add any external test for SPI, only loop back is sufficient to test the functionality.

b49020:
@VeshalaAnilKumar please address comments that have been long pending to address from Selvam and Copilot.

b49020:
Also, get rid of all the unnecessary code comments added by the AI agents.

core/drivers/spi/qcom/platform/lemans/qcom_geni_spi_config.c
```diff
@@ -0,0 +1,1383 @@
...
+		.clk_hz             = 19200000,
+		.se_clock_name      = "gcc_qupv3_wrap2_s6_clk",
+		.common_clocks_name = common_clocks_qup2,
+		.pin_groups         = qup2_se6_spi_pin_groups,
+		.pin_group_count    = ARRAY_SIZE(qup2_se6_spi_pin_groups),
+		.fw_image           = NULL,
+		.fw_image_size      = 0,
+	},
+#endif
+};
```
b49020: Please drop all other SPI configs, only keep the dTPM SPI on Lemans which we are interested in. Rest all other instances should be owned by Linux kernel.

core/arch/arm/plat-qcom/hoya/lemans/target.mk
```diff
@@ -1,9 +1,9 @@
 CFG_DRIVERS_CLK ?= y
 CFG_DRIVERS_QCOM_CLK ?= y
 
-CFG_QCOM_TLMM ?= y
-
 CFG_QCOM_DIAG_LOG ?= $(CFG_TEE_CORE_DEBUG)
+CFG_QCOM_GENI_SPI ?= y
+CFG_QUP2_SE2_SPI_EN ?= y
```
b49020: Is dTPM attached to this SPI instance on Lemans EVK?

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,1505 @@
...
+	/*
+	 * Hold the lock for the whole transfer, not just this setup. The
+	 * polling loop below consumes tx_buf/rx_buf and decrements
+	 * tx_rem_bytes/rx_rem_bytes through the FIFO helpers, and the
+	 * short-transfer check afterwards reads them, so unlocking before the
+	 * poll would leave a second caller free to overwrite that state
+	 * mid-transfer. Safe to span the poll now this is a mutex and not a
+	 * spinlock: no interrupts are disabled for the duration, and none of
+	 * the callees below take the lock themselves.
+	 */
```
b49020: Drop all this AI generated commentry.

b49020:
Please rebase this PR to tip of `qcom-next` too.

core/drivers/spi/qcom/platform/lemans/qcom_geni_spi_config.c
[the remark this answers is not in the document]
VeshalaAnilKumar: done

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,1494 @@
...
+
+	while (tx_free_words && qs->tx_rem_bytes) {
+		uint32_t word = 0;
+		unsigned int j = 0;
+
+		for (j = 0; j < qs->bytes_per_word &&
+		     qs->tx_rem_bytes; j++, qs->tx_rem_bytes--) {
+		     uint8_t b = qs->tx_buf ? *qs->tx_buf++ : 0;
+		     word |= (uint32_t)b << (8 * j);
+		}
```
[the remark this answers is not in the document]
VeshalaAnilKumar: done

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,1494 @@
...
+}
+
+/*
+ * Mux this SE's pads to the SPI function once, at init: build the TLMM
+ * pinctrl state and apply it. apply() claims pin ownership, so it runs a
+ * single time here and the pads stay owned/configured for the driver's
+ * lifetime. A platform that leaves pin_groups NULL is assumed to set up
+ * pinmux elsewhere. Idempotent: gated by qs->pin_state, so a repeat
+ * qup_spi_init() call is a no-op here (re-applying would be rejected by
+ * TLMM anyway, since ownership was never released).
```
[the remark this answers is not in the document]
VeshalaAnilKumar: done

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,1494 @@
...
+	 */
+	res = qup_spi_load_fw(qs);
+	if (res)
+		return res;
+
+	/*
+	 * This driver uses polling mode: CS assert/deassert and TX/RX
+	 * transfers all poll M_IRQ_STATUS from the calling thread (see
+	 * qup_spi_poll_m_cmd()) instead of registering an interrupt
+	 * handler, so qs->itr_num is not registered with the GIC here.
```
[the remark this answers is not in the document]
VeshalaAnilKumar: done

core/drivers/spi/qcom/qcom_geni_spi_test.c
VeshalaAnilKumar: done

core/drivers/spi/qcom/platform/lemans/qcom_geni_spi_config.c
VeshalaAnilKumar: sure.

but kept safety wise, if customer want's to test any another instance, they can enable easily instead of adding all the chagnes

core/arch/arm/plat-qcom/hoya/lemans/target.mk
VeshalaAnilKumar: yes

core/drivers/spi/qcom/qcom_geni_spi_test.c
```diff
@@ -0,0 +1,180 @@
...
+	TEE_Result res = TEE_SUCCESS;
+	int se_instance_id = 15;
+
+	if (!qup_spi_config_count) {
+		IMSG("QUP SPI test: no SE configured for this platform, skipping");
+		return TEE_SUCCESS;
+	}
+
+	/* instance id can vary based on use case */
+	res = qup_spi_init(&qs, se_instance_id);
```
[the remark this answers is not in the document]
VeshalaAnilKumar: done

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,1505 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2026, Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#include <assert.h>
+#include <drivers/clk.h>
+#include <drivers/clk_qcom.h>
+#include <drivers/qcom/tlmm/tlmm.h>
```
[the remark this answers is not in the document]
VeshalaAnilKumar: will put tlmm/clock has dependency chagnges

core/drivers/spi/qcom/sub.mk
```diff
@@ -0,0 +1,4 @@
+srcs-$(CFG_QCOM_GENI_SPI) += qcom_geni_spi.c
+srcs-$(CFG_QUP_SPI_TEST) += qcom_geni_spi_test.c
+
+subdirs-$(CFG_QCOM_GENI_SPI) += platform
```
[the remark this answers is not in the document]
VeshalaAnilKumar: done

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,1505 @@
...
+		 * Arm the watermark and do the initial fill directly:
+		 * M_IRQ_STATUS has not latched yet, so the polling loop
+		 * would see nothing to do on its first iteration otherwise.
+		 */
+		io_write32(qs->base + SE_GENI_TX_WATERMARK_REG, 1);
+		qup_spi_fifo_fill_tx(qs);
+	}
+
+	poll_res = qup_spi_poll_m_cmd(qs, M_CMD_DONE_EN, true,
+				       QUP_SPI_M_CMD_TIMEOUT_US);
```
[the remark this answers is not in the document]
VeshalaAnilKumar: done

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,1505 @@
...
+	TEE_Result res = TEE_SUCCESS;
+
+	assert(qs);
+
+	if (!qup_spi_get_platform_data(qup_spi_id, qs))
+		return TEE_ERROR_ITEM_NOT_FOUND;
+
+	qs->id = qup_spi_id;
+	qs->chip.ops = &qup_spi_ops;
+	mutex_init(&qs->lock);
```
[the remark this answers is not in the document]
VeshalaAnilKumar: done

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,1356 @@
...
+
+static const struct spi_ops qup_spi_ops = {
+	.configure = qup_spi_configure,
+	.start = qup_spi_start,
+	.txrx8 = qup_spi_txrx8,
+	.txrx16 = qup_spi_txrx16,
+	.end = qup_spi_end,
+	.flushfifo = qup_spi_flushfifo,
+};
+DECLARE_KEEP_PAGER(qup_spi_ops);
```
b49020: Drop these pager declarations anywhere in this patch-set, no support for it on Qcom platforms.

core/drivers/spi/qcom/qcom_geni_spi.c
b49020: Address this comment @VeshalaAnilKumar

core/drivers/spi/qcom/qcom_geni_spi.c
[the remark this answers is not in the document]
b49020: Address this too @VeshalaAnilKumar

b49020:
@VeshalaAnilKumar please split this single commit into 3 as follows:
1. Adding the SPI driver
2. Lemans SPI config
3. Last adding the loop back SPI test case.

core/drivers/spi/qcom/qcom_geni_spi.c
VeshalaAnilKumar: sure, will remove it

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,1356 @@
...
+	if (!num_pkts || num_pkts > TRANS_LEN_MSK) {
+		EMSG("QUP SPI %u: num_pkts=%zu out of range (1..%u)", qs->id,
+		     num_pkts, (unsigned int)TRANS_LEN_MSK);
+		return SPI_ERR_PKTCNT;
+	}
+
+	io_write32(qs->base + SE_SPI_TX_TRANS_LEN, wdat ? num_pkts : 0);
+	io_write32(qs->base + SE_SPI_RX_TRANS_LEN, rdat ? num_pkts : 0);
+
+	mutex_lock(&qs->lock);
```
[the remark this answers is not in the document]
VeshalaAnilKumar: done

core/drivers/spi/qcom/qcom_geni_spi_test.c
```diff
@@ -0,0 +1,128 @@
...
+	qs->bits_per_word = QUP_SPI_TEST_BITS_PER_WORD;
+	qs->mode = SPI_MODE0;
+	qs->cs = 0;
+	qs->cs_high = false;
+	qup_spi_set_loopback(qs, true);
+	qs->chip.ops->configure(&qs->chip);
+
+	res = qup_spi_test_run_txrx(qs, tx, rx, sizeof(tx));
+
+	qup_spi_set_loopback(qs, false);
```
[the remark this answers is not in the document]
VeshalaAnilKumar: done

core/drivers/spi/qcom/qcom_geni_spi.c
VeshalaAnilKumar: done

core/drivers/spi/qcom/platform/lemans/qcom_geni_spi_config.c
```diff
@@ -0,0 +1,225 @@
...
+#include <drivers/qcom/tlmm/tlmm.h>
+#include <drivers/qcom_geni_spi.h>
+#include <mm/core_mmu.h>
+#include <util.h>
+
+/*
+ * Lemans has a single SPI consumer: the dTPM on QUP2 SE2. Only that SE is
+ * configured; other platforms' SPI SEs belong to their own platform config.
+ */
+#ifdef CFG_QUP2_SE2_SPI_EN
```
b49020: Is this QUP SPI firmware data specific to QUP2 SE2? If not then remove it from this `#ifdef`.

core/arch/arm/plat-qcom/hoya/lemans/target.mk
```diff
@@ -4,6 +4,8 @@ CFG_DRIVERS_QCOM_CLK ?= y
 CFG_QCOM_TLMM ?= y
 
 CFG_QCOM_DIAG_LOG ?= $(CFG_TEE_CORE_DEBUG)
+CFG_QCOM_GENI_SPI ?= y
+CFG_QUP2_SE2_SPI_EN ?= y
```
b49020: I think this QUP2 SE2 config option doesn't serve any real purpose. Can we drop it?

b49020:
Changes looks good to me for `qcom-next` apart from 2 comments above unless copilot has more comments. Please raise an upstream PR with these comments incorporated.

core/drivers/spi/qcom/platform/lemans/qcom_geni_spi_config.c
VeshalaAnilKumar: No, qup-fw is generic.

core/include/drivers/qcom_geni_spi.h
```diff
@@ -0,0 +1,160 @@
...
+	struct clk *common_clks[QUP_SPI_MAX_COMMON_CLKS];
+	/* Number of entries populated in common_clks[] */
+	unsigned int num_common_clks;
+	/*
+	 * TLMM pin groups for this SE's pads, from the platform cfg. The
+	 * resolved pinctrl state is built and applied once by qup_spi_init().
+	 */
+	const struct tlmm_pin_group *pin_groups;
+	unsigned int pin_group_count;
+	struct pinctrl_state *pin_state;
```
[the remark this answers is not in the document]
VeshalaAnilKumar: done

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,1358 @@
...
+		if (qs->pin_state) {
+			tlmm_free_pin_state(qs->pin_state);
+			qs->pin_state = NULL;
+		}
+		qup_spi_clk_disable(qs);
+		return;
+	}
+	mutex_unlock(&qs->lock);
+
+	qs->started = true;
```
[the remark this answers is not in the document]
VeshalaAnilKumar: done

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,1358 @@
...
+		uint32_t src = rates[i];
+		uint32_t div = 0;
+		uint32_t sclk = 0;
+
+		if (src < speed_hz)
+			continue;
+
+		div = (src + speed_hz - 1) / speed_hz;
+		if (div > CLK_DIV_MAX)
+			continue;
```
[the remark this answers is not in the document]
VeshalaAnilKumar: done

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,1358 @@
...
+
+	if (qs->tx_fifo_depth < 4) {
+		EMSG("QUP SPI %u: implausible tx_fifo_depth=%u from SE_HW_PARAM_0 (clock gated?)",
+		     qs->id, qs->tx_fifo_depth);
+		qup_spi_clk_disable(qs);
+		return;
+	}
+	io_write32(qs->base + SE_GENI_RX_WATERMARK_REG, qs->tx_fifo_depth - 3);
+	io_write32(qs->base + SE_GENI_RX_RFR_WATERMARK_REG,
+		   qs->tx_fifo_depth - 2);
```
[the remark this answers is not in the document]
VeshalaAnilKumar: done

core/arch/arm/plat-qcom/hoya/lemans/target.mk
VeshalaAnilKumar: yes, will add changes in next patch

core/drivers/spi/qcom/qcom_geni_spi.c
b49020: This comment is still unaddressed.

core/drivers/spi/qcom/platform/lemans/qcom_geni_spi_config.c
```diff
@@ -0,0 +1,224 @@
...
+    0x45, 0x46, 0x47, 0x48, 0x49, 0x4a, 0x4b, 0x4c, 0x4d, 0x4e, 0x4f, 0x50, 0x51, 0x52, 0x53, 0x54,
+    0x55, 0x56, 0x57, 0x58, 0x59, 0x5a, 0x5b, 0x5c, 0x5d, 0x5e, 0x5f, 0x60, 0x61, 0x62, 0x63, 0x64,
+    0x65, 0x66, 0x67, 0x68, 0x69, 0x6a, 0x6b, 0x6c, 0x6d, 0x6e, 0x6f, 0x70, 0x71
+};
+
+/*
+ * Lemans has a single SPI consumer: the dTPM on QUP2 SE2. Only that SE is
+ * configured; other platforms' SPI SEs belong to their own platform config.
+ */
+#ifdef CFG_QUP2_SE2_SPI_EN
```
b49020: Please drop this macro check now since QUP2 SE2 will be enabled by default.

core/drivers/spi/qcom/platform/lemans/qcom_geni_spi_config.c
```diff
@@ -0,0 +1,224 @@
...
+		.func        = 1,
+		.pull        = TLMM_PULL_NONE,
+		.drive_ma    = 6,
+		.strong_pull = false,
+	},
+};
+#endif
+
+const struct qup_spi_platform_cfg qup_spi_config[] = {
+#ifdef CFG_QUP2_SE2_SPI_EN
```
b49020: Ditto here, drop this macro check.

core/drivers/spi/qcom/platform/lemans/qcom_geni_spi_config.c
VeshalaAnilKumar: sure

core/drivers/spi/qcom/platform/lemans/qcom_geni_spi_config.c
VeshalaAnilKumar: when we add multiple configs, it would required. 
as of now single instance is there, removing the macro

core/drivers/spi/qcom/qcom_geni_spi.c
```diff
@@ -0,0 +1,1370 @@
...
+
+static const struct spi_ops qup_spi_ops = {
+	.configure = qup_spi_configure,
+	.start = qup_spi_start,
+	.txrx8 = qup_spi_txrx8,
+	.txrx16 = qup_spi_txrx16,
+	.end = qup_spi_end,
+	.flushfifo = qup_spi_flushfifo,
+};
+DECLARE_KEEP_PAGER(qup_spi_ops);
```
b49020: This declare pager is still here, please drop

core/drivers/spi/qcom/qcom_geni_spi.c
VeshalaAnilKumar: sorry, updated now

b49020:
Merging into `qcom-next` now, since verified on Lemans EVK the SPI loop back test passes:

```
I/TC: 
I/TC: OP-TEE version: optee.os.0.0-00018-5-g741f505b5 (gcc version 15.2.0 (Ubuntu 15.2.0-16ubuntu1)) #1 Tue Sep 22 07:26:48 UTC 2026 aarch64
I/TC: WARNING: This OP-TEE configuration might be insecure!
I/TC: WARNING: Please check https://optee.readthedocs.io/en/latest/architecture/porting_guidelines.html
I/TC: Primary CPU initializing
I/TC: TLMM: base=0xf000000, 149 GPIOs
I/TC: QUP SPI 15: fw loaded (protocol=1 fw_version=0xb02 cfg_version=0x9)
I/TC: QUP SPI 15: initialized (irq 616)
I/TC: QUP SPI 15: applied 2 pin group(s)
I/TC: QUP SPI 15: loopback test: PASS (16 bytes)
I/TC: Platform Qualcomm: Flavor lemans
I/TC: Primary CPU switching to normal world boot
```

Thanks @VeshalaAnilKumar for the rework. Please now go ahead with upstream PR.

PR #40, opened 2026-08-11
Nord Clock driver Support

nnunna94:


ldts:
**Architectural review (recommendations).**

Two robustness issues worth addressing before a consumer lands:

1. `qcom_qup_clk_enable()` sets the vote bit (`io_setbits32(vote, BIT(dom->vote_bit))`) and then returns `TEE_ERROR_TIMEOUT` if the CBCR never reports on. The clk framework keeps the enable-count at 0 on failure, so no later `clk_disable()` will clear that vote — the branch stays voted-on with no owner. Recommend clearing the vote bit before the timeout return.

2. `qcom_qup_clocks_register()` publishes the `qcom_qup_clks` global (via `calloc`) *before* the per-clk alloc/register loop completes. If any `clk_alloc`/`clk_register` fails mid-loop, unfilled entries keep `domain == NULL`, and `qcom_clk_get_by_name()` iterates `bsp->n_domains` dereferencing `qcom_qup_clks[i].domain->name`. Also, on partial failure the global is left non-NULL, so a re-run hits the `if (qcom_qup_clks) return SUCCESS` guard and reports success without finishing registration. Recommend building into temp storage and publishing only after full success.

nnunna94:
> **Architectural review (recommendations).**
> 
> Two robustness issues worth addressing before a consumer lands:
> 
> 1. `qcom_qup_clk_enable()` sets the vote bit (`io_setbits32(vote, BIT(dom->vote_bit))`) and then returns `TEE_ERROR_TIMEOUT` if the CBCR never reports on. The clk framework keeps the enable-count at 0 on failure, so no later `clk_disable()` will clear that vote — the branch stays voted-on with no owner. Recommend clearing the vote bit before the timeout return.
> 2. `qcom_qup_clocks_register()` publishes the `qcom_qup_clks` global (via `calloc`) _before_ the per-clk alloc/register loop completes. If any `clk_alloc`/`clk_register` fails mid-loop, unfilled entries keep `domain == NULL`, and `qcom_clk_get_by_name()` iterates `bsp->n_domains` dereferencing `qcom_qup_clks[i].domain->name`. Also, on partial failure the global is left non-NULL, so a re-run hits the `if (qcom_qup_clks) return SUCCESS` guard and reports success without finishing registration. Recommend building into temp storage and publishing only after full success.

This PR needs rebasing and refactoring once PR26 lands @ldts . Hence this is in Draft.

PR #41, opened 2026-08-12
plat-qcom: nord: add PIL bring-up for HPASS, SOCCP and Turing DSPs

taniyadas20:
Add the clock and PAS bring-up support required to load the additional DSP images on the Nord platform:

 - clock-qcom: add a Lucid-OLE PLL enable helper. Lucid-OLE reuses the Lucid-EVO register layout but packs PLL_L_VAL as an 8-bit L value plus separate 8-bit process-cal and ring-osc-cal fields, and requires the PLL_TEST_CTL* trim registers to be programmed, so it gets its own config struct and enable path.
 - clk_qcom: add QCOM_CLKS_{SOCCP,HPASS0,HPASS1,HPASS2,TURING2,TURING3} clock groups and route them through the PAS enable path.
 - pas_data: add PAS IDs for TURING2/3, HPASS0-2 and SOCCP.
 - nord: add the per-platform clock-qcom-pas.c and clock_group_qcom.h describing the PLL and clock-group configuration for these subsystems.
 - arch/target config: add the register base/size definitions needed by the new bring-up code.


core/drivers/clk/qcom/platform/nord/clock-qcom-pas.c
spalepuqual: We need to checkin soccp.c as well to enable SoCCP PIL

PR #42, opened 2026-08-14
Add support for Hardware key manager v1 based In-line Crypto Engine

harshaldev27:
Add support for HWKM v1 based In-line Crypto Engine Psuedo-TA


core/drivers/crypto/qcom/hwkm/transaction.c
```diff
@@ -392,89 +396,142 @@ static int hwkm_fifo_wait(vaddr_t base, uint32_t reg, uint32_t mask)
...
+	if (!ctx->ice_base)
+		ctx->ice_base = (vaddr_t)phys_to_virt(HWKM_ICE_BASE,
+						      MEM_AREA_IO_SEC,
+						      HWKM_ICE_SIZE);
+
+	/* ICE clocks and power are handled by Linux, so we need to re-configure
+	 * the hardware before submitting any new transactions.
+	 */
+	rc = hwkm_ice_configure(ctx->ice_base);
+	if (!rc)
```
ldts: **Functional bug: inverted return check disables every ICE-slave transaction.**

```c
rc = hwkm_ice_configure(ctx->ice_base);
if (!rc)
        return rc;
```

`hwkm_ice_configure()` returns `HWKM_SUCCESS` (== 0) on success. So on **success** this returns immediately *without* ever calling `run_fifo_transaction()`, and it only runs the transaction when configure **failed** (against an unconfigured ICE). Every ICE-slave key program/clear silently no-ops. The check should be `if (rc) return rc;`.

harshaldev27: Thanks, resolved.

b49020:
@harshaldev27 can you rather only pick HWKM patches from @qc-azarrabi PR which the ICE feature depends upon? This can untangle GPCE and ICE PRs.

harshaldev27:
> Harshal Dev (@harshaldev27) can you rather only pick HWKM patches from Amirreza Zarrabi (@qc-azarrabi) PR which the ICE feature depends upon? This can untangle GPCE and ICE PRs.

Done, only 3 commits from his PR related to HWKM re-factoring and GPCE slave addition are now kept.

PR #45, opened 2026-08-24
Pil: nord, shikra and cacao

pawarai123:


zelvam95:
What is the goal of this PR @pawarai123? Can we please split it per target? Separate PR for Nord/Shikra & Cacao? This PR currently has ~7000 lines of code.

Also, If this PR is still not tested, could you please consider moving it to draft? (Only if its not tested/not yet ready for review)

PR #47, opened 2026-08-24
drivers: qcom: rpmh: fix MSGID length field encoding

shvm-ap:
The RPMh command MSGID encodes a MSG_LENGTH field describing the
payload length, in bytes, of the command. This field was hardcoded
to 1 which is leading to unpredictable behavior on the
AOP side (including the command never being acknowledged, causing timeouts).

Changing it to 8 (the correct length for the single 32-bit data
word every RPMh command carries), introducing macros MSGID_MSG_LENGTH_VALUE,
MSGID_WRITE, MSGID_READ so the expected encoding is explicit.

Using MSGID_WRITE since it's a write command.

### DIFF
--- a/core/drivers/qcom/rpmh/rpmh_hal.c
+++ b/core/drivers/qcom/rpmh/rpmh_hal.c
@@ -230,9 +230,9 @@ enum hal_status hal_rpmh_write_cmd(enum rsc_drv_id drv_id,
 	cmd_base = tcs_base + TCS_CMD_BASE_OFFSET + (cmd_idx * TCS_CMD_STRIDE);
 
 	msgid = 0;
-	msgid |= SHIFT_U32(0, MSGID_READ_OR_WRITE_SHIFT);
+	msgid |= SHIFT_U32(MSGID_WRITE, MSGID_READ_OR_WRITE_SHIFT);
 	msgid |= SHIFT_U32((completion ? 1 : 0), MSGID_RES_REQ_SHIFT);
-	msgid |= SHIFT_U32(1, MSGID_MSG_LENGTH_SHIFT);
+	msgid |= SHIFT_U32(MSGID_MSG_LENGTH_VALUE, MSGID_MSG_LENGTH_SHIFT);
 
 	slave_id = (addr >> 16) & 0x7;
 	offset = addr & 0xFFFF;
--- a/core/drivers/qcom/rpmh/rpmh_hal.h
+++ b/core/drivers/qcom/rpmh/rpmh_hal.h
@@ -53,6 +53,12 @@ enum hal_status {
 #define MSGID_RES_REQ_SHIFT		0x8
 #define MSGID_MSG_LENGTH_SHIFT		0x0
 
+#define MSGID_READ			0x0
+#define MSGID_WRITE			0x1
+
+/* Each RPMh command transfers a single 32-bit word (encoded as 8 bytes). */
+#define MSGID_MSG_LENGTH_VALUE		0x8
+
 #define ADDR_SLV_ID_SHIFT		0x10
 #define ADDR_OFFSET_SHIFT		0x0
 


IDineshChoudhary (approved):
This is required changes for client's rpmh voting, 0x1 -> maps to undefined behavior in hardware

zelvam95:
> This is required changes for client's rpmh voting, 0x1 -> maps to undefined behavior in hardware

Thanks for raising this! As discussed over DM, strangely enough, this didn't affect fuse blowing in Lemans / Kodiak. I tested multiple OEM Spare/SHK fuses across boards and it was working. 

That said, given it can lead to undefined behaviour, we can prioritize to fix this in upstream as well and it should be easy to get it upstream'd given its a small change.

**Can you please plan to push this PR in upstream @shvm-ap? We can directly review it there and try to get it merged.**

shvm-ap:
Sure @zelvam95.

core/drivers/qcom/rpmh/rpmh_hal.c
```diff
@@ -232,7 +232,7 @@ enum hal_status hal_rpmh_write_cmd(enum rsc_drv_id drv_id,
 	msgid = 0;
 	msgid |= SHIFT_U32(0, MSGID_READ_OR_WRITE_SHIFT);
```
IDineshChoudhary: now that we are updating it via macro possible to follow same here?

IDineshChoudhary (changes requested):

core/drivers/qcom/rpmh/rpmh_hal.c
```diff
@@ -230,9 +230,9 @@ enum hal_status hal_rpmh_write_cmd(enum rsc_drv_id drv_id,
 	cmd_base = tcs_base + TCS_CMD_BASE_OFFSET + (cmd_idx * TCS_CMD_STRIDE);
 
 	msgid = 0;
-	msgid |= SHIFT_U32(0, MSGID_READ_OR_WRITE_SHIFT);
+	msgid |= SHIFT_U32(MSGID_READ, MSGID_READ_OR_WRITE_SHIFT);
```
IDineshChoudhary: this is also a problem, it should be write not read, but I will go with validations and approve (something to be aware).

IDineshChoudhary (approved):

core/drivers/qcom/rpmh/rpmh_hal.c
zelvam95: > this is also a problem, it should be write not read, but I will go with validations and approve (something to be aware).

How is it in TF-A? If its not OR'd in TF-A, we could just align to it?

IDineshChoudhary: it's WRITE one in TF-A, with current set of bugs in here RPMH voting should have never worked @zelvam95

IDineshChoudhary: surprisingly with READ how clocks team validations are working - that's further interesting (serious doubts on validations now, proxy votes are saving us is all I can say at this time).

IDineshChoudhary: for TF-A reference you can see this: https://github.com/qualcomm-linux/trusted-firmware-a/blob/qcom-next/drivers/qti/rpmh/rpmh_client.c#L89

IDineshChoudhary: @shvm-ap as discussed offline, let's go with the changes whatever is working and validated by clocks team - raise a gerrit for same.

zelvam95: Seems to work for clocks team as well as you said; 
Yes its probably due to proxy voting being done earlier.

Can we please fix that as well Dinesh? I think it'd make sense to fix the issues rather than waiting for issues to be identified later. Can we request clocks team for quick test with the updated change? Would strongly recommend we fix any identified BUGs.

zelvam95: Would also request a review of the CMD_DB/RPMH driver in upstream to identify any other potential issues in OP-TEE^^ @IDineshChoudhary.

zelvam95: One other update; Looks like the rail voting is required specifically for Kodiak & not Lemans; But yes, still a mystery how it worked in Kodiak;

>Kodiak requires MX voltage rail workaround for QFPROM fuse blowing
$(call force,CFG_QFPROM_MX_RAIL_WA,y)

IDineshChoudhary (approved):

zelvam95:
@shvm-ap / @IDineshChoudhary, If further review is happening upstream, do you think this can be merged to qcom-next?

IDineshChoudhary:
@shvm-ap can you add details on unit testing as well rail and bandwidth voting is verified.

zelvam95:
Hi @shvm-ap,

Upstream PR: https://github.com/OP-TEE/optee_os/pull/7958
Qcom-next PR: https://github.com/qualcomm-linux/optee_os/pull/57

As we synced offline, the AMC Enable/completion tracking part is fixed as part of the above PR. Can we cherry-pick the above commits and do unit-test together? If it passes, we could merge these PRs together in qcom-next & upstream.

IDineshChoudhary (approved):

IDineshChoudhary (approved):
Looks good to me, functional changes for RPMH infrastructures to work - Approving.

zelvam95 (approved):

shvm-ap:
@ldts @b49020 please review. Added fix tag as well.

b49020:
Based on approvals, will merge this now.

PR #49, opened 2026-08-25
pta: qcom: pas: nord: register HPASS0-2 and NSPSS0-3 PAS subsystems

deepaknasir:
Register the PAS bring-up for Nord's subsystems HPASS & NSPSS.
- hpass0-2: program the HPASS reset-EVB and TCSR EVB-select registers for each HPASS instance from its own base offset and TCSR stride index.
- nspss0-3: program the NSPSS reset-EVB register for each CDSP instance.
- subsys.c: register all subsystems (HPASS0-2, NSPSS0-3) with their PAS IDs, base/size and clock groups.

### DIFF
--- a/core/pta/qcom/pas/platform/nord/hpass0.c
+++ b/core/pta/qcom/pas/platform/nord/hpass0.c
@@ -1,0 +1,81 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2026, Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+
+#include <io.h>
+#include <platform_config.h>
+#include <resource_table.h>
+#include <stdint.h>
+#include <string.h>
+
+#include "hpass0.h"
+
+/*
+ * QDSP6 boot registers for each HPS instance. The PUB block
+ * holds the reset/boot FSM registers; the TCSR block holds the EVB select. 
+ */
+#define HPASS_BASE_OFFSET				(0xC00000)
+#define HPASS_TCSR_REGS_REG_OFFSET		(0x1580000)
+
+#define HPASS_QDSP6SS_RST_EVB			(0x10)
+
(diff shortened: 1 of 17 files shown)


ldts:
**Architectural review (recommendation).**

The nord PAS bring-up isn't self-contained and has a latent bug, though it's currently dormant (`CFG_QCOM_PAS_PTA` is enabled only for kodiak/lemans, not nord, so this code isn't compiled yet). Worth fixing before it's turned on:

- `subsys.c` references many symbols that aren't defined anywhere in the tree or this PR: `CDSP_0..3_BASE/SIZE`, `HPASS_0..2_BASE/SIZE`, `PAS_ID_TURING2/3`, `PAS_ID_HPASS1/2` (`pas_data.h` defines only up to `PAS_ID_TURING1`), and `QCOM_CLKS_TURING2/3`, `QCOM_CLKS_HPASS0/1/2`. These need corresponding additions to `platform_config.h` / `pas_data.h` / the `qcom_clk_group` enum.

- In `hpass1.c`/`hpass2.c`, `HPASS_EFUSE_Q6SS_EVB_SEL = HPASS_TCSR_REGS_REG_OFFSET − HPASS_BASE_OFFSET + ...` is negative (hpass1 base `0x2000000`, hpass2 `0x2100000` both exceed the TCSR offset `0x1580000`), so `io_write32(base + HPASS_EFUSE_Q6SS_EVB_SEL, 0)` targets a wild offset. Only `hpass0` (base `0xC00000`) stays positive, which suggests the per-instance `HPASS_BASE_OFFSET` values for instances 1/2 were mis-derived.

Since `hpass0-2` / `nspss0-3` are 7 near-identical copy-pasted files differing only in a few constants, a single parameterized implementation would both fix the underflow class of bug and cut the maintenance surface.

b49020:

